PR #1, opened 2021-05-07
Revert "converted prop types to safer React.ComponentPropsWithRef, op…

nilueps:
switched up the component props strategy based on advice for building a react component librairy found here: https://react-typescript-cheatsheet.netlify.app/docs/advanced/intro

Also switched from `import [module], { ...module } from 'react'` to `import * as React from 'react`


nilueps:
nvm I messed up my commit history!

PR #13, opened 2021-06-21
bugfix: case view restricts parties options based on party type

nilueps:
Should work. But SSR forces page refresh on case type choice. Maybe a better solution would filter the parties client-side?

### DIFF
--- a/api2/pkg/parties/parties/client.go
+++ b/api2/pkg/parties/parties/client.go
@@ -19,11 +19,21 @@ func NewClient(basePath string) *Client {
 	}
 }
 
+type ListOptions struct {
+	PartyType string `json:"partyTypeId" bson:"partyTypeId"`
+}
+
 func (c *Client) List(ctx context.Context, listOptions ListOptions) (*PartyList, error) {
 	req, err := http.NewRequest("GET", c.basePath+"/apis/v1/parties", nil)
 	if err != nil {
 		return nil, err
 	}
+	//req = req.WithContext(ctx)
+	qry := req.URL.Query()
+	if len(listOptions.PartyType) > 0 {
+		qry.Set("partyTypeId", listOptions.PartyType)
+	}
+	req.URL.RawQuery = qry.Encode()
 	req.Header.Set("Accept", "application/json")
 	res, err := http.DefaultClient.Do(req)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


api2/pkg/parties/parties/client.go
```diff
@@ -19,11 +19,21 @@ func NewClient(basePath string) *Client {
...
+type ListOptions struct {
+	PartyType string `json:"partyTypeId" bson:"partyTypeId"`
+}
+
 func (c *Client) List(ctx context.Context, listOptions ListOptions) (*PartyList, error) {
 	req, err := http.NewRequest("GET", c.basePath+"/apis/v1/parties", nil)
 	if err != nil {
 		return nil, err
 	}
+	//req = req.WithContext(ctx)
```
ludydoo, line 31: Don't remove this ;)

api2/pkg/webapp/cases.go
```diff
@@ -127,19 +127,29 @@ func (h *Handler) NewCase(w http.ResponseWriter, req *http.Request) {
...
-	})
-
 	if err := g.Wait(); err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
 
 	qry := req.URL.Query()
+	caseTypeId := qry.Get("caseTypeId")
+	var partyTypeId string
```
ludydoo, line 137: partyTypeID := ""

api2/pkg/webapp/templates/casenew.gohtml
```diff
@@ -31,7 +31,7 @@
                     </div>
                     <div class="form-floating mb-3">
                         <select id="party" class="form-select" name="partyId">
-                            {{if .Parties}}
+                            {{if and .Parties}}
```
ludydoo, line 34: no need for and

api2/pkg/webapp/templates/casenew.gohtml
```diff
@@ -56,6 +56,14 @@
             </div>
         </div>
     </div>
+    <script>
+        const caseTypeSelect = document.querySelector("#case-type");
+        caseTypeSelect.onchange = (event) => {
+            const caseTypeId = event.target.value;
+            if (caseTypeId === "") return;
```
ludydoo, line 63: use braces for all ifs

```
if (caseTypeId === ""){
  return
}
```

ludydoo (approved):

PR #14, opened 2021-06-21
Cor 11 referral from caseview

nilueps:
Implementing this revealed another issue we haven't addressed. Namely, that the referral UI should allow the user to select a  `CaseType` only if that `CaseType` has the proper `PartyTypeID` for the given `Party`.

I already created another [ticket](https://dteam-nrc.atlassian.net/browse/COR-26) for this and assigned it to myself


nilueps:
On second though, I'll fix this on this branch and make another PR

PR #15, opened 2021-06-22
Beneficiary relationship dropdown

shinroo:
- implement dropdown that filters based on user search term
- add full text index


api2/go.mod
```diff
@@ -3,24 +3,23 @@ module github.com/nrc-no/core-kafka
 go 1.16
```
ludydoo, line 3: Why are there changes to go.mod ?

api2/pkg/parties/parties/store.go
```diff
@@ -7,17 +7,17 @@ import (
 )
 
 type Store struct {
-	collection *mongo.Collection
+	Collection *mongo.Collection
```
ludydoo, line 10: I don't think this should be exported... 
```suggestion
	collection *mongo.Collection
```

api2/pkg/relationshipparties/rest.go
```diff
@@ -0,0 +1,49 @@
...
+	ctx := req.Context()
+
+	qry := req.URL.Query()
+
+	pickPartyOptions := PickPartyOptions{
+		PartyTypeID: qry.Get("partyTypeId"),
+		SearchParam: qry.Get("searchParam"),
+	}
+
+	ret, err := h.partiesStore.FilteredList(ctx, pickPartyOptions)
```
ludydoo, line 34: Can't we use list ?
```suggestion
	ret, err := h.partiesStore.List(ctx, listOptions)
```

api2/pkg/relationshipparties/static.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"context"
+	"github.com/nrc-no/core-kafka/pkg/parties/attributes"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+func Init(ctx context.Context, partiesStore *PartiesStore) error {
+	if _, err := partiesStore.store.Collection.Indexes().CreateOne(ctx, mongo.IndexModel{
```
ludydoo, line 11: I would move this to individuals/static.go or parties/static.go (this is not really a concern of relationshipparties, but rather of individuals)

api2/pkg/server/server.go
```diff
@@ -37,41 +38,44 @@ import (
...
-	HttpServer              *http.Server
-	TeamStore               *teams.Store
-	TeamHandler             *teams.Handler
-	TeamClient              *teams.Client
-	MembershipHandler       *memberships.Handler
-	MembershipStore         *memberships.Store
-	membershipsClient       *memberships.Client
-	KeycloakClient          *keycloak.Client
-	SessionManager          sessionmanager.Store
+	MongoClient                *mongo.Client
```
ludydoo, line 41: How come are there so many changes ?

api2/pkg/server/server.go
```diff
@@ -437,41 +451,42 @@ func (c CompletedOptions) New(ctx context.Context) *Server {
 	}
 
 	srv := &Server{
```
ludydoo, line 453: How come are there so many changes here too ?

api2/pkg/webapp/templates/individual.gohtml
```diff
@@ -139,13 +143,29 @@
...
+        // Handles the search box for the parties dropdown
+        let partySearchSubject = new Subject()
+        let partySearch$ = partySearchSubject.asObservable().pipe(
+            debounceTime(200),
+            switchMap((searchParam) => {
+                const searchParams = new URLSearchParams({
+                    "partyTypeId": getSecondPartyType(),
+                    "searchParam": searchParam
+                })
+                const ajaxUrl = "{{.BasePath}}" + "/apis/v1/relationshipparties/picker?" + searchParams.toString()
```
ludydoo, line 161:
```suggestion
                const ajaxUrl = "/apis/v1/relationshipparties/picker?" + searchParams.toString()
```

api2/pkg/webapp/individuals.go
```diff
@@ -154,14 +161,18 @@ func (h *Handler) Individual(w http.ResponseWriter, req *http.Request) {
...
+		"IsNew":              id == "new",
+		"Individual":         b,
+		"Parties":            bList,
+		"PartyTypes":         partyTypes,
+		"RelationshipTypes":  relationshipTypes,
+		"Relationships":      relationshipsForIndividual,
+		"Attributes":         attrs,
+		"Cases":              displayCases,
+		"CaseTypes":          ctList,
+		"BasePath":           h.relationshipPartiesClient.BasePath,
```
ludydoo, line 173: We can remove this
```suggestion

```

api2/pkg/webapp/templates/individual.gohtml
```diff
@@ -240,18 +260,127 @@
...
-                let partySelector = document.getElementById('relationshipSecondPartySelector');
+                let partySelector = document.getElementById('partySelector');
                 let relationshipType = typeSelector.value;
-                let party = partySelector.value;
+                let party = partySelector.getAttribute("data-party-id");
                 addRelationship(relationshipType, party);
             };
 
+            // Add listener for party selector item click
+            Array.from(document.getElementsByClassName('partySelectorItem')).forEach(e => {
```
ludydoo, line 326: Do we need Array.from?

ludydoo (changes requested):

api2/go.mod
shinroo: When I ran it in Goland it automatically changed go.mod

api2/pkg/parties/parties/store.go
shinroo: I need it to be exported to be able to wrap it in the relationshipparties store

api2/pkg/webapp/templates/individual.gohtml
shinroo: Yes, because getElementsByClassName returns an HTMLCollection which cannot directly be iterated over with Array methods. To be able to use .forEach on it, I needed to change it to an Array first.

See here: [JS: iterating over result of getElementsByClassName using Array.forEach](https://stackoverflow.com/questions/3871547/js-iterating-over-result-of-getelementsbyclassname-using-array-foreach)

api2/pkg/server/server.go
shinroo: I'm not sure to be honest, this came about when I merged with main

api2/go.mod
shinroo: Ran `go mod tidy` as suggested by @ludydoo on slack

ludydoo (changes requested):

PR #19, opened 2021-06-22
COR-2 user authentication

ludydoo:
This is a big PR.

Some code was merged into main beforehand, but I'm gonna consider its part of this PR 

1. An `auth` handler that handles `login`, `logout` and `callback` handler for oidc
2. An `auth` middleware that validates if a user is logged in. If not, redirects to the login page.
3. Possibility of using the `session` that contains the access token, or pass the access token in the `Authorization` header
4. Fix tests to log in as admin/admin user automatically
5. Each client must forward the access token when talking to the api (loopback)
6. Create dummy admin/admin user on startup


api2/pkg/auth/handler.go
```diff
@@ -33,7 +40,12 @@ func init() {
 	gob.Register(&Profile{})
 }
 
-func NewHandler(ctx context.Context, issuerURL, clientID, clientSecret, redirectURL string, store sessionmanager.Store) (*Handler, error) {
+func NewHandler(
```
shinroo, line 43: We should introduce a consistent rule for how to format params to functions like this. Is it possible to make that a team-wide editor setting as well @ludydoo ?

shinroo (commented):
Hey, took a browse through.

I noticed we apply this in most client methods:
```go
auth.SetAuthorizationHeader(ctx, req)
```
But not in the relationshipparties client. Should this change be made there as well?

ludydoo:
@shinroo good find! yes that would be necessary, I can add it

api2/pkg/auth/handler.go
nilueps: I think Go comes with it's own formatting tool `go fmt` that should handle most cases like these. In goland you can format a file with it using `Ctrl+Alt+Shift+F`.

nilueps (approved):

PR #22, opened 2021-06-22
improve case view

ludydoo:


ludydoo:
Screenshot from 2021-06-22 19-10-46
Screenshot from 2021-06-22 19-10-37

shinroo (commented):
I think having the case status as a tag might be better, as this would make it easier to introduce other statuses like `inReview` along the lines of what Abraham showed us today

nilueps:
I agree, a named tag is less ambiguous. But maybe this is something we should ask Cassie about?

ludydoo:
I think that a case could be either done or not done. After that they might have different statuses (think Kanban board). Issues can be in progress, todo, in review, etc. But they are also either done or not done

I imagine that each team would have their own statuses

nilueps:
In that case, maybe at least the shape should change, like 'X' for "not done" and 'V' (a check-mark) for "done". Relying only on color is bad for color-blind folks

nilueps (commented):
From the [BS Icons](https://icons.getbootstrap.com/#icons) there's 'x-circle' and 'check-circle'

nilueps (commented):
.

nilueps (changes requested):
err, still gettinh the hang of this

nilueps:
I'm gonna merge this because I need to work on these files and I anticipate conflicts.

Please let me know if there's a better solution.

PR #26, opened 2021-06-23
COR-36 redirect users to the case view after create/update

ludydoo:


api2/pkg/cases/cases/rest.go
```diff
@@ -105,8 +105,8 @@ func (h *Handler) Put(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	r.PartyID = payload.PartyID
-	r.CaseTypeID = payload.CaseTypeID
+	// r.PartyID = payload.PartyID
```
shinroo, line 108: Can we delete these commented lines?

api2/pkg/webapp/cases.go
```diff
@@ -196,8 +196,10 @@ func (h *Handler) PostCase(ctx context.Context, id string, w http.ResponseWriter
 	done := req.Form.Get("done")
 	parentId := req.Form.Get("parentId")
 
+	var kase *cases.Case
```
shinroo, line 199: Why do we define kase and err here, rather than using the := notation inside the if/else?

ludydoo: Because the kase might be either populated from Create or Update (its a shared var)

api2/pkg/cases/cases/rest.go
ludydoo: Yes!

PR #27, opened 2021-06-23
COR-34 redirect users when token is expired

ludydoo:


api2/pkg/auth/middleware.go
```diff
@@ -23,47 +24,59 @@ func (h *Handler) Authenticate() func(handler http.Handler) http.Handler {
...
-					return
 				}
+			}
 
-				jkws, err := h.KeycloakClient.GetPublicKeys()
-				if err != nil {
-					doNext(req, false)
-					return
-				}
+			if len(accessToken) == 0 {
```
shinroo, line 37: Since we have `isInStore`, could this be `!isInStore` instead?

ludydoo: Not really, because the access token might come from the store or from the Authorization header...

PR #30, opened 2021-06-23
COR-22: select casetype team

shinroo:
1. Add team ID to case type struct
2. On case type creation show dropdown for team


ludydoo (approved):
Looking good!!

PR #33, opened 2021-06-23
COR-41: Refactor relationship type rules

shinroo:
- Renamed party type fields to party type ids
- Had to make some small changes to Case code to get the tests to run, but the case tests still have 2 failing cases @ludydoo 
- Checked the features in the UI and it seems to work, despite the failing Case tests

### DIFF
--- a/api2/pkg/webapp/casetypes.go
+++ b/api2/pkg/webapp/casetypes.go
@@ -41,6 +41,49 @@ func (h *Handler) CaseTypes(w http.ResponseWriter, req *http.Request) {
 	}
 }
 
+func (h *Handler) PostCaseType(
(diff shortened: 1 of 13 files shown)


api2/pkg/webapp/casetypes.go
```diff
@@ -41,6 +41,49 @@ func (h *Handler) CaseTypes(w http.ResponseWriter, req *http.Request) {
 	}
 }
 
+func (h *Handler) PostCaseType(
```
shinroo, line 44: @senyaoh I had to move this function above the others and fix the missing } problem!

ludydoo (approved):

PR #36, opened 2021-06-23
Cor 32 bootstrap cypress

nilueps:
Instructions in README, should be fairly straightforward. Next step is writing the actual tests :)


nilueps:
Just realized my files didn't get commited

shinroo (approved):
Looks good to me, was able to install cypress and run the tests according to the README

PR #37, opened 2021-06-23
COR-43-swapped-case-status-symbols

nilueps:
They were just flipped


shinroo (approved):
Good move for accessibility!

PR #38, opened 2021-06-23
COR-23: Show case teams on new case and case page

senyaoh:
Show case teams on new case and case page

### DIFF
--- a/api2/pkg/webapp/cases.go
+++ b/api2/pkg/webapp/cases.go
@@ -7,6 +7,7 @@ import (
 	"github.com/nrc-no/core-kafka/pkg/cases/cases"
 	"github.com/nrc-no/core-kafka/pkg/cases/casetypes"
 	"github.com/nrc-no/core-kafka/pkg/parties/parties"
+	"github.com/nrc-no/core-kafka/pkg/teams"
 	"golang.org/x/sync/errgroup"
 	"net/http"
 )
@@ -64,6 +65,8 @@ func (h *Handler) Case(w http.ResponseWriter, req *http.Request) {
 
 	var referrals *cases.CaseList
 
+	var team *teams.Team
+
 	g, waitCtx := errgroup.WithContext(ctx)
 
 	g.Go(func() error {
@@ -85,6 +88,15 @@ func (h *Handler) Case(w http.ResponseWriter, req *http.Request) {
 	if err := g.Wait(); err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
+	}
+
+	if id != "new" {
+		teamRes, err := h.teamClient.Get(ctx, kase.TeamID)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		team = teamRes
 	}
 
 	party, err := h.partyClient.Get(ctx, kase.PartyID)
@@ -124,6 +136,7 @@ func (h *Handler) Case(w http.ResponseWriter, req *http.Request) {
 		"Parties":          partyList,
 		"ReferralCaseType": referralCaseType,
 		"Referrals":        referrals,
+		"Team":             team,
 	}); err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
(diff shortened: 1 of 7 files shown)


api2/pkg/webapp/cases.go
```diff
@@ -87,6 +90,15 @@ func (h *Handler) Case(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	if id != "new" {
+		teamRes, err := h.teamClient.Get(ctx, kase.TeamID)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
```
shinroo, line 96: @senyaoh @ludydoo this implies that if we have no case team selected, we will get an error page. Is this the correct behaviour?

nilueps (changes requested):
Looks good to me.
I would just consider renaming `TeamList` to `Teams`.
`TeamList` is arguably clearer, but the plural form is what we've been using so far so I would suggest sticking to that for now and consider renaming all such variables to the `...List` form in another PR.

api2/pkg/webapp/cases.go
nilueps: In that case, I believe the method would return `(nil, nil)`, so no error, normally

api2/pkg/webapp/templates/casenew.gohtml
```diff
@@ -29,6 +29,22 @@
...
+                                    <option {{if eq .ID $.TeamID}}selected{{end}} value="{{.ID}}">{{.Name}}</option>
+                                {{end}}
+                            {{else}}
+                                <option selected disabled>No parties defined</option>
+                            {{end}}
+                        </select>
+                        <label for="team">Select Team</label>
+                    </div>
+                    <input id="teamId" name="teamId" type="hidden" value="{{.TeamID}}"/>
+
```
ludydoo, line 47: In this section, the Team has to be bound to the CaseType. 
Since CaseTypes are owned by teams

2 options:
1. Make the Team readonly. When the user changes the CaseType, then set the team using javascript
2. When the user changes the team, filter the CaseType list to only show the case types for that team

ludydoo (changes requested):

api2/pkg/webapp/templates/casenew.gohtml
shinroo: Hey @ludydoo those requirements are met by the current implementation.

We asked Nico for advice, since this page redirects the user each time the case type is changed (it adds the caseTypeId as a URL param and the backend sends it back down when the template is re-rendered).

We followed a similar approach with the TeamID, we used JS to map the caseTypeId to a teamId and then we attach that to the URL sent with the redirect.

The following shows how it works:
https://user-images.githubusercontent.com/15915453/123253597-ed63a400-d4ed-11eb-8ad1-6f2712cd0871.mp4

ludydoo: I see, then I think there is no need for the teamId request parameter, as the teamId is already inferred from the caseType ?

shinroo:
```javascript
const caseTypeTeamMap = {
            {{range .CaseTypes.Items}}
            "{{.ID}}": "{{.TeamID}}",
            {{end}}
        }

        const caseTypeSelect = document.querySelector("#case-type");
        caseTypeSelect.onchange = (event) => {
            const caseTypeId = event.target.value;
            if (caseTypeId === "") {
                return;
            }
            window.location.href = `/cases/new?caseTypeId=${caseTypeId}&teamId=${caseTypeTeamMap[caseTypeId]}`
        }
```

shinroo: You are right @ludydoo, we could actually just get the team info server side and send it down!

nilueps (commented):
Removed the unused TeamList, otherwise A-OK!

nilueps (approved):

api2/pkg/webapp/cases.go
```diff
@@ -157,24 +168,35 @@ func (h *Handler) NewCase(w http.ResponseWriter, req *http.Request) {
 	qry := req.URL.Query()
 	caseTypeID := qry.Get("caseTypeId")
 	partyTypeID := ""
+	teamID := ""
 	for _, caseType := range caseTypes.Items {
 		if caseType.ID == caseTypeID {
 			partyTypeID = caseType.PartyTypeID
+			teamID = caseType.TeamID
```
ludydoo, line 175: minor: add a break
```suggestion
			teamID = caseType.TeamID
			break
```

ludydoo (changes requested):

api2/pkg/webapp/cases.go
ludydoo: minor correction, feel free to approve & merge after that

PR #39, opened 2021-06-24
COR-13: modify individual List to only include individuals and not all parties

shinroo:
- New default behaviour of List is to filter on partyTypeIds looking for individuals
- If ListOptions is passed in, the behaviour is to `and` the list options filter with the individual partyTypeIds filter

### DIFF
--- a/api2/pkg/individuals/store.go
+++ b/api2/pkg/individuals/store.go
@@ -58,12 +58,32 @@ func (s *Store) Get(ctx context.Context, ID string) (*Individual, error) {
 
 func (s *Store) List(ctx context.Context, listOptions ListOptions) (*IndividualList, error) {
 
-	filter := bson.M{}
+	filter := bson.M{
+		"partyTypeIds": bson.M{
+			"$elemMatch": bson.M{
+				"$eq": partytypes.IndividualPartyType.ID,
+			},
+		},
+	}
 
 	if len(listOptions.PartyTypeIDs) > 0 {
-		filter["partyTypeIds"] = bson.M{
-			"$all": listOptions.PartyTypeIDs,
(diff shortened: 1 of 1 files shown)


api2/pkg/individuals/store.go
```diff
@@ -58,12 +58,32 @@ func (s *Store) Get(ctx context.Context, ID string) (*Individual, error) {
...
+				bson.M{
+					"partyTypeIds": bson.M{
+						"$all": listOptions.PartyTypeIDs,
+					},
+				},
+				bson.M{
+					"partyTypeIds": bson.M{
+						"$elemMatch": bson.M{
+							"$eq": partytypes.IndividualPartyType.ID,
+						},
```
ludydoo, line 81:
```suggestion
                includesIndividualType := false
                for _, partyTypeID := listOptions.PartyTypeIds {
                  if partyTypeID == partytypes.IndividualPartyType.ID {
                    includesIndividualType = true
                  } 
                }
                if !inclduesIndividualType {
                  listOptions.PartyTypeIds = append(listOptions.PartyTypeIDs, partytypes.IndividualPartyType.ID)
                }
                filter["partyTypeIds"] = bson.M{
			"$all": listOptions.PartyTypeIDs
	         }
	         
```

ludydoo (changes requested):

api2/pkg/individuals/store.go
ludydoo: Here, you can just add IndividualPartyTypeID if it is not already in listOptions.PartyTypeIDs

shinroo: Changes made!

PR #40, opened 2021-06-24
COR-15: only show relevant attributes

shinroo:
- On attribute API, move from static subject type to using party type
- Use hardcoded static individual type in the subject type dropdown on the attribute definition page
- Filter the list of attributes in the backend to retrieve only those who apply to a beneficiary

### DIFF
--- a/api2/pkg/webapp/templates/attribute.gohtml
+++ b/api2/pkg/webapp/templates/attribute.gohtml
@@ -45,7 +45,9 @@
                                 <select id="subject-type-input"
                                         name="partyTypes"
                                         class="form-select">
-                                    <option>Individual</option>
+                                    {{range .PartyTypes.Items}}
+                                    <option value="{{.ID}}">{{.Name}}</option>
+                                    {{end}}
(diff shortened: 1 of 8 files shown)


api2/pkg/webapp/templates/attribute.gohtml
```diff
@@ -118,20 +120,24 @@
...
                 valueType: "",
-                partyTypes: "individual",
+                partyTypes: ["{{(index .PartyTypes.Items 0).ID}}"],
                 isPersonallyIdentifiableInformation: false,
                 translations: []
             },
             partyTypes: [
-                {value: "individual", name: "Individual"}
+                {{range .PartyTypes.Items}}
+                {value: "{{.ID}}", name: "{{.Name}}"}
```
ludydoo, line 139: You will need a comma at the end of this line
```suggestion
                {value: "{{.ID}}", name: "{{.Name}}"},
```

ludydoo (changes requested):
Looks good! minor change

PR #59, opened 2021-06-29
COR-78 memberships unmarshalqueryparams

ludydoo:


shinroo (approved):
Nice, this will make working with queries a much more pleasant experience!

PR #60, opened 2021-06-29
COR-79: relationship picker bug

shinroo:
- Add handler for picking relationships in the webapp
- Add new endpoint for picking relationships to the webapp
- Ajax call from individuals now uses webapp endpoint and not api endpoint


api/.idea/workspace.xml
```diff
@@ -1,13 +1,22 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="ChangeListManager">
-    <list default="true" id="400da93d-09a9-4b62-b409-2409f1a925fa" name="Default Changelist" comment="Renamed api2 directory to api, changed project name from core-kafka to core" />
+    <list default="true" id="400da93d-09a9-4b62-b409-2409f1a925fa" name="Default Changelist" comment="Renamed api2 directory to api, changed project name from core-kafka to core">
```
ludydoo, line 4: Not sure this file should be changed..

api/pkg/apps/webapp/relationshippartypicker.go
```diff
@@ -0,0 +1,34 @@
+package webapp
+
+import (
+	"encoding/json"
+	"github.com/nrc-no/core/pkg/apps/iam"
+	"net/http"
+)
+
+func (h *Server) PickParty(w http.ResponseWriter, req *http.Request){
```
ludydoo, line 9: Can you rename to PickRelationshipParty?
```suggestion
func (h *Server) PickRelationshipParty(w http.ResponseWriter, req *http.Request){
```

api/pkg/apps/webapp/server.go
```diff
@@ -113,6 +113,7 @@ func NewServer(
 	router.Path("/settings/authclients/{id}").HandlerFunc(h.AuthClient)
 	router.Path("/settings/authclients/{id}/newsecret").HandlerFunc(h.AuthClientNewSecret)
 	router.Path("/settings/authclients/{id}/delete").HandlerFunc(h.DeleteAuthClient)
+	router.Path("/relationships/pickparty").HandlerFunc(h.PickParty)
```
ludydoo, line 116:
```suggestion
	router.Path("/relationships/pickparty").HandlerFunc(h.PickRelationshipParty)
```

ludydoo (changes requested):
Looks good! Just minor changes for naming and a file that I believe should not be changed

ludydoo (approved):

PR #61, opened 2021-06-29
COR-67 introducing Page Object Pattern

keth:
Introducing Page Object Pattern


shinroo (commented):
Thanks @keth I have learned something new today 👍

ludydoo (approved):

nilueps (approved):

PR #66, opened 2021-06-29
COR-12 filter cases by team/status/casetype

ludydoo:


nilueps (changes requested):
I'm getting a `context canceled` error when trying to create a referral.
Also, only open cases are displayed, is this intentional?

keth (approved):

ludydoo:
Nice find @nilueps

PR #68, opened 2021-06-29
COR-94-case-types-edit-bug

nilueps:


keth (approved):

api/pkg/apps/webapp/templates/casetype.gohtml
```diff
@@ -57,6 +57,10 @@
                         </select>
                         <label for="team">Select Team</label>
                     </div>
+                    {{if .CaseType}}
+                        <input type="hidden" name="partyTypeId" value="{{.CaseType.PartyTypeID}}">
```
ludydoo, line 61: Hm, I think the teamId / partyTypeId should be readonly. I don't think we need to have them as hidden fields. What is the reason for the change?

nilueps: It seemed the disabled `<select>` don't pass their values to the request, so a hidden input was necessary

nilueps: Or maybe they we're being passed as empty strings? In any case, the original value was getting lost

ludydoo:
Maybe we don’t render the value? I would check that instead of adding a
hidden field

Le mer. 30 juin 2021 à 14:02, Nicolas Epstein ***@***.***> a
écrit :

api/pkg/apps/webapp/templates/casetype.gohtml
ludydoo: Exactly, so we can put the  CaseTypeID <select> field as `readonly` when this is an existing case instead of `disabled` so that the value does not get lost. We can use a single field to represent that value instead of using 2 different fields in the codebase, I think its a rendering problem

PR #69, opened 2021-06-29
Cor 93 add user notifications

nilueps:
Notifications are working now for most successful actions.
For some reason, notifs don't show up on Attribute updating, only creating, if one of you can take a look, maybe a fresh pair of eyes will spot the culprit.


keth:
> Notifications are working now for most successful actions.
> For some reason, notifs don't show up on Attribute updating, only creating, if one of you can take a look, maybe a fresh pair of eyes will spot the culprit.

Is sounds like this is something we could catch with Unit-Tests

keth (approved):

api/pkg/apps/webapp/attributes.go
```diff
@@ -67,18 +68,18 @@ func (h *Server) Attribute(w http.ResponseWriter, req *http.Request) {
...
-	a, err := iamClient.Attributes().Get(ctx, id)
+	attribute, err := iamClient.Attributes().Get(ctx, id)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
 
 	if req.Method == "POST" {
-		h.PostAttribute(ctx, a, w, req)
+		h.PostAttribute(ctx, attribute, w, req)
```
ludydoo, line 78: return here ?

api/pkg/apps/webapp/templates/casetype.gohtml
```diff
@@ -57,6 +57,10 @@
                         </select>
                         <label for="team">Select Team</label>
                     </div>
+                    {{if .CaseType}}
```
ludydoo, line 60: whats that for?

api/pkg/apps/webapp/attributes.go
nilueps: good catch ;)

api/pkg/apps/webapp/templates/casetype.gohtml
nilueps: values from the disabled selects above we're not getting through

ludydoo:
That doesn’t seem right

Le mer. 30 juin 2021 à 14:05, Nicolas Epstein ***@***.***> a
écrit :

api/pkg/apps/webapp/templates/casetype.gohtml
nilueps: maybe it was something to do with the values not being passed along properly, I didn't take the time to investigate to be honest.

looking into it now

nilueps:
Ok it turns out disabled input elements are not submitted with forms as I suspected
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled

ludydoo:
Then read only attribute ?

Le mer. 30 juin 2021 à 14:16, Nicolas Epstein ***@***.***> a
écrit :

nilueps:
yeah

nilueps:
rememeber, we're using the same template for new casetypes as for viewing created ones. so if `.CaseType` if present, we chose to disabled those inputs

PR #70, opened 2021-06-29
COR-24 Show referral creator

nilueps:
Records the referring user's first and last names for the moment. Ideally would record a unique identifier.

### DIFF
--- a/api/pkg/apps/webapp/cases.go
+++ b/api/pkg/apps/webapp/cases.go
@@ -264,6 +264,18 @@ func (h *Server) PostCase(ctx context.Context, id string, w http.ResponseWriter,
 	var kase *cms.Case
 	if id == "" {
 		var err error
+		// record user id of creator if referral
+		// TODO get actual unique identifier rather than name
+		creatorId := ""
+		if len(parentId) > 0 {
+			profileIntf := h.sessionManager.Get(ctx, "profile")
+			profile, ok := profileIntf.(*Claims)
+			if !ok {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+			creatorId = fmt.Sprintf("%s %s", profile.GivenName, profile.FamilyName)
+		}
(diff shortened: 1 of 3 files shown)


api/pkg/apps/webapp/cases.go
```diff
@@ -264,13 +264,26 @@ func (h *Server) PostCase(ctx context.Context, id string, w http.ResponseWriter,
 	var kase *cms.Case
 	if id == "" {
 		var err error
+		// record user id of creator if referral
+		// TODO get actual unique identifier rather than name
```
keth, line 268: It would be nice to create a jira-ticket and the issueKey to the comment (if this is something we want TODO)

keth (approved):

api/pkg/apps/webapp/cases.go
```diff
@@ -264,13 +264,26 @@ func (h *Server) PostCase(ctx context.Context, id string, w http.ResponseWriter,
...
+		creatorId := ""
+		if len(parentId) > 0 {
+			profileIntf := h.sessionManager.Get(ctx, "profile")
+			profile, ok := profileIntf.(*Claims)
+			if !ok {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+			creatorId = fmt.Sprintf("%s %s", profile.GivenName, profile.FamilyName)
+		}
```
ludydoo, line 278: You can get the author from ctx.Value("subject").(string). See `comment_post.go`
```suggestion
		creatorId := ctx.Value("subject").(string)
```

ludydoo (changes requested):

api/pkg/apps/webapp/cases.go
ludydoo: We can use the `context.Value(ctx, "subject").(string)` to get the user that made that request

ludydoo (changes requested):

ludydoo (approved):

PR #78, opened 2021-07-01
Cor 66 e2e db seed

nilueps:
fixed small bug


ludydoo (changes requested):
Already fixed by COR-107

PR #85, opened 2021-07-05
COR-122 Case template config convention

nilueps:


api/pkg/apps/cms/types.go
```diff
@@ -53,3 +54,17 @@ type Comment struct {
...
+// https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema
+
+type CaseTemplate struct {
+	// FormElements is an ordered list of the elements found in the form
+	FormElements []CaseTemplateFormElement `json:"formElements" bson:"formElements"`
+}
+
+type CaseTemplateFormElement struct {
+	Type       string                 `json:"type" bson:"type"`
+	Attributes map[string]interface{} `json:"attributes" bson:"attributes"`
```
ludydoo, line 68: I think `Attributes` and `Validation` should be structs as well...

ludydoo (changes requested):

api/pkg/apps/cms/types.go
nilueps: The thing is, attributes and validation fields (and field types) will vary depending on the form element. Should we have structs for each element?

ludydoo: @nilueps I don't think so, we can put all the available attributes in the same struct, and have validation rules to make sure the payload is in the right format

ludydoo (approved):

PR #87, opened 2021-07-05
Cor 119 render new case template

nilueps:
This PR includes some static code to add a demo case template to all the builtin casetypes.

### DIFF
--- a/api/pkg/apps/cms/types.go
+++ b/api/pkg/apps/cms/types.go
@@ -7,14 +7,31 @@ import (
 )
 
 type Case struct {
-	ID          string `json:"id" bson:"id"`
-	CaseTypeID  string `json:"caseTypeId" bson:"caseTypeId"`
-	PartyID     string `json:"partyId" bson:"partyId"`
-	Description string `json:"description" bson:"description"`
-	Done        bool   `json:"done" bson:"done"`
-	ParentID    string `json:"parentId" bson:"parentId"`
-	TeamID      string `json:"teamId" bson:"teamId"`
-	CreatorID   string `json:"creatorId" bson:"creatorId"`
+	ID          string        `json:"id" bson:"id"`
+	CaseTypeID  string        `json:"caseTypeId" bson:"caseTypeId"`
+	PartyID     string        `json:"partyId" bson:"partyId"`
+	Description string        `json:"description" bson:"description"`
+	Done        bool          `json:"done" bson:"done"`
+	ParentID    string        `json:"parentId" bson:"parentId"`
+	TeamID      string        `json:"teamId" bson:"teamId"`
+	CreatorID   string        `json:"creatorId" bson:"creatorId"`
+	FormData    *CaseTemplate `json:"formData" bson:"formData"`
+}
+
+func (c *Case) UnmarshalFormData(values url.Values, caseTemplate *CaseTemplate) error {
+	c.CaseTypeID = values.Get("caseTypeId")
+	c.PartyID = values.Get("partyId")
+	c.Description = values.Get("description")
+	c.Done = values.Get("done") == "on"
+	c.ParentID = values.Get("parentId")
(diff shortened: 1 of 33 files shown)


api/pkg/apps/cms/types.go
```diff
@@ -33,6 +37,19 @@ func (c *CaseType) String() string {
 	return c.Name
 }
 
+func (c *CaseType) UnmarshalFormData(values url.Values) error {
+	c.Name = values.Get("name")
+	c.PartyTypeID = values.Get("partyTypeId")
+	c.TeamID = values.Get("teamId")
+	templateString := values.Get("template")
```
ludydoo, line 44: Can we unmarshal the whole CaseTemplate object rather than unmarshaling the array of CaseTemplateFormElements? That would allow us to put some more config in the textbox, rather than just the inner array of formElements

eg `json.Unmarshal([]byte(templateString), &c.CaseTemplate)`

api/pkg/apps/seeder/static.go
```diff
@@ -1,22 +1,34 @@
...
+func caseTemplate(templateString string) cms.CaseTemplate {
+	var formElements []cms.CaseTemplateFormElement
+	if err := json.Unmarshal([]byte(templateString), &formElements); err != nil {
+		panic(err)
+	}
+	return cms.CaseTemplate{
+		FormElements: formElements,
+	}
+}
+
```
ludydoo, line 31: This is duplicated

ludydoo (changes requested):

api/pkg/apps/cms/types.go
nilueps: Yeah, I just thought it was cumbersome to have to wrap every template in a `formElements` containing field, but if we're going to have more root-level items, I agree with you

api/pkg/apps/cms/types.go
```diff
@@ -63,22 +80,36 @@ type CaseTemplate struct {
...
 }
 
+func NewCaseTemplate(templateString string) (CaseTemplate, error) {
+	var ct CaseTemplate
+	if err := json.Unmarshal([]byte(templateString), &ct); err != nil {
+		return ct, err
+	}
+	return ct, nil
+}
+
```
ludydoo, line 90: This is exactly the same signature as the vanilla json.Unmarshal, so this method is not needed
You can replace the calls by json.Unmarshal

api/pkg/apps/seeder/static.go
```diff
@@ -88,12 +89,15 @@ var (
 	NairobiResponseTeam = team("814fc372-08a6-4e6b-809b-30ebb51cb268", "Nairobi Response Team")
 	NairobiICLATeam     = team("80606eb4-b53a-4fda-be12-e9806e11d44a", "Nairobi ICLA Team")
 
+	// Case Templates
+	Legal, _ = cms.NewCaseTemplate(`{"formElements": [{"type":"dropdown","attributes":{"label":"Legal status","id":"legalStatus","description":"What is the beneficiary's current legal status?","options":["Citizen","Permanent resident","Accepted refugee","Asylum seeker","Undetermined"]},"validations":{"required":true}},{"type":"checkbox","attributes":{"label":"Qualified services","id":"qualifiedServices","description":"What services does the beneficiary qualify for?","checkboxOptions":[{"label":"Councelling"},{"label":"Representation"},{"label":"Arbitration"}]},"validations":{"required":true}},{"type":"textarea","attributes":{"label":"Notes","id":"notes","description":"Additional informations, observations, concerns, etc.","placeholder":"Type something here..."}}]}`)
```
ludydoo, line 93: Can we replace with the actual structs ?
```
Legal = cms.CaseTemplate{...}
```

ludydoo (changes requested):

ludydoo (approved):

PR #88, opened 2021-07-06
Cor 132 save filled case template

nilueps:
This PR implements saving form data for a form generate through our new case templating system.

When creating a template, the form elements are associated to a FormData map that binds the html input names of the rendered form to objects in the following example structure:

```
formData: {
        legalStatus: {                   // a dropdown
            type: 'string',
            value: [
                '0'
            ]
        },
        notes: {                            // a textarea
            type: 'string',
            value: [
                'Example string'
            ]
        },
        qualifiedServices: {          // checkboxes
            type: 'bool',
            value: [
                '0',
                '2'
            ]
        }
    }
```

### DIFF
--- a/api/pkg/apps/cms/types.go
+++ b/api/pkg/apps/cms/types.go
@@ -42,11 +42,11 @@ func (c *CaseType) UnmarshalFormData(values url.Values) error {
 	c.PartyTypeID = values.Get("partyTypeId")
 	c.TeamID = values.Get("teamId")
 	templateString := values.Get("template")
-	var formElements []CaseTemplateFormElement
-	if err := json.Unmarshal([]byte(templateString), &formElements); err != nil {
+	template, err := NewCaseTemplate(templateString)
+	if err != nil {
 		return err
 	}
-	c.Template = CaseTemplate{FormElements: formElements}
+	c.Template = template
 	return nil
 }
 
@@ -78,6 +78,14 @@ type CommentList struct {
 type CaseTemplate struct {
 	// FormElements is an ordered list of the elements found in the form
 	FormElements []CaseTemplateFormElement `json:"formElements" bson:"formElements"`
+}
+
+func NewCaseTemplate(templateString string) (CaseTemplate, error) {
+	var ct CaseTemplate
+	if err := json.Unmarshal([]byte(templateString), &ct); err != nil {
+		return ct, err
+	}
+	return ct, nil
 }
 
 type CaseTemplateFormElement struct {
--- a/api/pkg/apps/cms/types.go
+++ b/api/pkg/apps/cms/types.go
@@ -49,19 +49,6 @@ func (c *CaseType) UnmarshalFormData(values url.Values) error {
 	return nil
 }
 
-func parseTemplate(str string) (CaseTemplate, error) {
-	var template []CaseTemplateFormElement
-
-	err := json.Unmarshal([]byte(str), &template)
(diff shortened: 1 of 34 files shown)


api/pkg/apps/cms/types.go
```diff
@@ -61,6 +79,29 @@ type CommentList struct {
...
+// NewCaseTemplate constructs a new case template from a JSON template string and creates a CaseFormData data structure
+// that with hold the data corresponding to the fields described in the template
+func NewCaseTemplate(templateString string) (*CaseTemplate, error) {
+	var ct CaseTemplate
+	if err := json.Unmarshal([]byte(templateString), &ct); err != nil {
+		return &ct, err
+	}
+	ct.FormData = ct.ToCaseFormData()
+	return &ct, nil
+}
```
ludydoo, line 96: This method should not be needed. We can simply use the vanilla `json.Unmarshal/json.Marshal`
The `values` of a `CaseTemplate` shoudl be stored in the `caseTemplate...Attribute.Value`

api/pkg/apps/seeder/static.go
```diff
@@ -88,12 +89,15 @@ var (
 	NairobiResponseTeam = team("814fc372-08a6-4e6b-809b-30ebb51cb268", "Nairobi Response Team")
 	NairobiICLATeam     = team("80606eb4-b53a-4fda-be12-e9806e11d44a", "Nairobi ICLA Team")
 
+	// Case Templates
+	Legal, _ = cms.NewCaseTemplate(`{"formElements": [{"type":"dropdown","attributes":{"label":"Legal status","id":"legalStatus","description":"What is the beneficiary's current legal status?","options":["Citizen","Permanent resident","Accepted refugee","Asylum seeker","Undetermined"]},"validations":{"required":true}},{"type":"checkbox","attributes":{"label":"Qualified services","id":"qualifiedServices","description":"What services does the beneficiary qualify for?","checkboxOptions":[{"label":"Councelling"},{"label":"Representation"},{"label":"Arbitration"}]},"validations":{"required":true}},{"type":"textarea","attributes":{"label":"Notes","id":"notes","description":"Additional informations, observations, concerns, etc.","placeholder":"Type something here..."}}]}`)
```
ludydoo, line 93: Can we use the struct here instead of json?
```
 Legal := cms.CaseTemplate{
   FormElements: []cms.FormElement{
   ...
    }
 }
```

api/pkg/apps/webapp/cases.go
```diff
@@ -369,12 +376,27 @@ func (h *Server) PostCase(ctx context.Context, id string, w http.ResponseWriter,
...
-	parentId := req.Form.Get("parentId")
-	teamId := req.Form.Get("teamId")
+	values := req.Form
+
+	caseTypeId := values.Get("caseTypeId")
+	partyId := values.Get("partyId")
+	description := values.Get("description")
+	done := values.Get("done")
+	parentId := values.Get("parentId")
+	teamId := values.Get("teamId")
```
ludydoo, line 386: Do you mind putting this in 
```
func (c *CaseType) UnmarshalQueryParams(values url.Values) error {
...
}
```
?

ludydoo: Actually I think the method already exists..

ludydoo (changes requested):

ludydoo (approved):

PR #94, opened 2021-07-07
COR-136 make form readonly on case view

nilueps:


api/pkg/apps/webapp/templates/form_element.gohtml
```diff
@@ -6,14 +6,16 @@
                     class="form-select"
                     aria-label="Default select example"
                     {{if .Attributes.Multiple}}multiple{{end}}
-                    {{if .Validation.Required}}required{{end}}>
+                    {{if .Validation.Required}}required{{end}}
+                    {{if .Attributes.Value}}disabled{{end}}
```
ludydoo, line 10: Value might not be there if required=false, so disabled should be true when case is not a new case

api/pkg/apps/webapp/templates/form_element.gohtml
```diff
@@ -32,6 +34,7 @@
                    type="checkbox"
                    {{if .Required}}required{{end}}
                     {{range $v := $.Attributes.Value }}{{if eq $istr $v}}checked{{end}}{{end}}
+                    {{if $.Attributes.Value}}disabled{{end}}
```
ludydoo, line 37: here too

api/pkg/apps/webapp/templates/form_element.gohtml
```diff
@@ -55,7 +62,9 @@
                    class="form-control"
                    type="text"
                    placeholder="{{.Attributes.Placeholder}}"
-                   {{if .Validation.Required}}required{{end}}>
+                   {{if .Validation.Required}}required{{end}}
+                    {{if .Attributes.Value}}disabled{{end}}
```
ludydoo, line 66: here too

api/pkg/apps/webapp/templates/form_element.gohtml
```diff
@@ -41,9 +44,13 @@
             <textarea id="{{.Attributes.ID}}"
                       name="{{.Attributes.ID}}"
                       class="form-control"
-                      style="height: 150px"
+                      style="height: 150px; white-space: pre"
                       placeholder="{{.Attributes.Placeholder}}"
-                      {{if .Validation.Required}}required{{end}}>{{if .Attributes.Value}}{{index .Attributes.Value 0}}{{end}}</textarea>
+                      {{if .Validation.Required}}required{{end}}
+                    {{if .Attributes.Value}}disabled{{end}}
```
ludydoo, line 50: here too

ludydoo (changes requested):

nilueps (commented):
it should be fine now

ludydoo:
Lol I still don't understand what the `filled` property means?

nilueps:
For the moment, we use the same template (`form_element.gohtml`) for rendering a new case form as for rendering a filled-out case form on the case view.

When rendering on the case view we want to disable/make readonly all the input controls. My initial approach was to disable a control if the associated `FormElement` had a value.

You pointed out that some elements might not be required and therefore may not have any value, but should still be disabled on the CaseView.

My solution was to introduce a new `Filled bool` field on the form elements which gets set when a form is successfully submitted. Which is then used in the template thus:
`{{ or .Attributes.Value (and (not .Attributes.Value) $filled)}}disabled{{end}}`

Without this extra datapoint, how do we distinguish unrequired input fields with no value that should be disabled from those that shouldn't?

Perhaps `Filled` is a poorly chosen name, though, maybe `Processed`, or `Set` would be better?

nilueps:
@ludydoo you can re-review this

nilueps:
@shinroo the `Dict` template function I introduce in this PR could be useful for COR-160 if you choose to save the script as a template. Embedded templates (ie `{{template "something" [pipeline]}}`) can only be invoked with one "pipeline argument", the `Dict` function allows you to create a map so that you can pass in as many objects as needed: `{{template "something" Dict "a" .A "b" .B}}`

shinroo (changes requested):
Hey Nico, the addition of the Dict is very interesting indeed.

Code looks good in general.

Only issue I face is that I cannot start the server on this branch, I get the following error:
```
# github.com/nrc-no/core/pkg/apps/webapp
../pkg/apps/webapp/renderer.go:30:5: cannot use &RendererFactory{} (type *RendererFactory) as type RenderInterface in assignment:
	*RendererFactory does not implement RenderInterface (wrong type for Notifications method)
		have Notifications() ([]*sessionmanager.Notification, error)
		want Notifications() []*sessionmanager.Notification
../pkg/apps/webapp/renderer.go:56:25: cannot use f (type *RendererFactory) as type RenderInterface in argument to WithRenderInterface:
	*RendererFactory does not implement RenderInterface (wrong type for Notifications method)
		have Notifications() ([]*sessionmanager.Notification, error)
		want Notifications() []*sessionmanager.Notification
../pkg/apps/webapp/renderer.go:72:43: cannot use renderer (type *Renderer) as type RenderInterface in argument to WithRenderInterface:
	*Renderer does not implement RenderInterface (wrong type for Notifications method)
		have Notifications() ([]*sessionmanager.Notification, error)
		want Notifications() []*sessionmanager.Notification
../pkg/apps/webapp/renderer.go:85:5: cannot use &Renderer{} (type *Renderer) as type RenderInterface in assignment:
	*Renderer does not implement RenderInterface (wrong type for Notifications method)
		have Notifications() ([]*sessionmanager.Notification, error)
		want Notifications() []*sessionmanager.Notification
```

Is this a problem on my side?

shinroo (approved):
It works perfectly on my machine now!

PR #96, opened 2021-07-07
COR-92: Rewrite e2e test with page object pattern

senyaoh:


nilueps:
Looking good so far!

You may want to merge main into this branch so that you can test the latest version; the CaseType and Case flows have changed a bit.

Feel free to delete the old `create-edit-*.spec.js` files I created once you have a working analog with the page object pattern.

e2e/cypress/helpers/index.js
```diff
@@ -1,16 +1,102 @@
-
-
 export const Urls = {
     CASES_URL: '/cases',
     NEW_CASE_URL: '/cases/new',
     NEW_ATTRIBUTE_URL: '/settings/attributes/new',
     NEW_CASETYPE_URL: '/settings/casetypes/new',
     NEW_INDIVIDUAL_URL: '/individuals/new',
-    NEW_PARTYTYPE_URL: '/settings/partytypes/new',
+    NEW_ENTITYTYPE_URL: '/settings/partytypes/new',
```
nilueps, line 7: Can we stick to `PARTY` ? I know it says Entity in the UI, but we intend to change it.

e2e/cypress/pages/newCasetype.page.js
```diff
@@ -7,36 +7,34 @@ const TEMPLATE = '[data-cy=template]';
...
+        return this;
+    };
+
+    selectTeam = (value) => {
+        cy.get(TEAM).select(value);
+        return this;
+    };
+
+    typeTemplate = (value) => {
+        cy.get(TEMPLATE).type(value);
```
nilueps, line 32: can you change this so that cypress injects the whole text directly? Currently, the text is being inputed character by character, which is quite long.

Perhaps in your page object you could have something like

```suggestion
        cy.get(TEMPLATE).value = value
```

nilueps:
Looking good :)

Just a couple more things:
- Adapt the Case tests to the new template system. Bear in mind, the built-in Cases are currently broken. Perhaps the best solution would be to make sure the Case test always runs right after the CaseType test so that we can ensure the Case form displays the proper fields.
- Add a test for the "search and add member" feature on a Team's page
    - Visit a Team's page
    - Type a valid name in the search box: ex: "Diddley"
    - Add it and verify it's been added
- Add "edit" tests for Individual, Attribute, Party, CaseType and Relationship
- After every "create" or "edit" a checks to make sure the given data was properly saved
- Remove the `main-pages.spec.js` we don't need it anymore

ludydoo (approved):

ludydoo (changes requested):
See @nilueps comments

keth (approved):

nilueps:
made a couple changes; see latest commits. You may want to double check your tests as some of them aren't passing with the latest changes.
If you're sick of this ticket though, go ahead and merge 😅  we can open tickets to update the tests

nilueps (approved):

PR #97, opened 2021-07-09
COR-116 Test untested backend api (iam)

nilueps:
All tests passing
TODO: Party Search test
TODO: cms API tests

About the migration of internal methods from public to private. While fiddling with test coverage, I was under the mistaken impression that only public methods appeared in the report, so I flipped all the methods that weren't part of the public API. Turns out the coverage report checks internals too, which makes sense.
In any case, given `iam`'s current package structure, I think the change is pertinent anyway.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 *.DS_Store
 api/certs/cert.pem
 api/certs/key.pem
+*/coverage.out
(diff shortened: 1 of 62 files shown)


nilueps:
About the migration of internal methods from public to private. While fiddling with test coverage, I was under the mistaken impression that only public methods appeared in the report, so I flipped all the methods that weren't part of the public API. Turns out the coverage report checks internals too, which makes sense.
In any case, given `iam`'s current package structure, I think the change makes sense.

ludydoo (approved):

PR #101, opened 2021-07-12
Cor 124 cms api validation

nilueps:
So this probably needs some work, what do you guys think?


shinroo (commented):
Looks good to me, like the popover and changes to the response messages!

nilueps:
@ludydoo can you check this?

ludydoo (approved):

api/pkg/apps/cms/case_post.go
```diff
@@ -15,6 +16,19 @@ func (s *Server) PostCase(w http.ResponseWriter, req *http.Request) {
...
+
+	errList := ValidateCase(kase, &validation.Path{})
+	if len(errList) > 0 {
+		status := validation.Status{
+			Status:  validation.Failure,
+			Code:    http.StatusUnprocessableEntity,
+			Message: "invalid case type",
+			Errors:  errList,
+		}
+		s.JSON(w, status.Code, status)
```
ludydoo, line 28: Same there here to extract the method

api/pkg/apps/cms/case_validation.go
```diff
@@ -0,0 +1,20 @@
...
+
+func ValidateCase(kase *Case, path *validation.Path) validation.ErrorList {
+	errList := validation.ErrorList{}
+
+	// Validate form elements
+	for _, elem := range kase.FormData.FormElements {
+		if elem.Validation.Required && len(elem.Attributes.Value) == 0 {
+			errList = append(errList, validation.Required(path.Child(elem.Attributes.ID), fmt.Sprintf("%s is required", elem.Attributes.ID)))
+		}
+		// TODO implement validation for specific input controls (email, date, etc)
```
ludydoo, line 16: Jira issue ?

api/pkg/apps/cms/casetype_validation.go
```diff
@@ -0,0 +1,21 @@
...
+
+func ValidateCaseType(caseType *CaseType, path *validation.Path) validation.ErrorList {
+	errList := validation.ErrorList{}
+
+	if len(caseType.Name) == 0 {
+		errList = append(errList, validation.Required(path.Child("name"), "name is required"))
+	} else if !validation.IsValidAlpha(caseType.Name) {
+		errList = append(errList, validation.Invalid(path.Child("name"), caseType.Name, "name should only contain letters and spaces"))
+	}
+	if len(caseType.PartyTypeID) == 0 {
```
ludydoo, line 13: We should also make sure that all UUIDS are actually valid UUIDs
```
func IsValidUUID(str string) bool {
_, err := uuid.FromString(str);
return err == nil
}
```

api/pkg/apps/webapp/casetypes.go
```diff
@@ -17,89 +17,29 @@ func (s *Server) CaseTypes(w http.ResponseWriter, req *http.Request) {
...
-		s.PostCaseType(ctx, &cms.CaseType{}, w, req)
-		return
-	}
-
 	caseTypes, err := cmsClient.CaseTypes().List(ctx, cms.CaseTypeListOptions{})
-	if err != nil {
-		s.Error(w, err)
+	if errResponse(w, err) {
 		return
 	}
```
ludydoo, line 23: You'll probably have merge conflicts on master for these

api/pkg/apps/webapp/casetypes.go
```diff
@@ -132,60 +70,120 @@ func (s *Server) CaseType(w http.ResponseWriter, req *http.Request) {
...
+	req *http.Request,
+	partyTypes *iam.PartyTypeList,
+	teamsData *iam.TeamList,
+) {
+	cmsClient := h.CMSClient(ctx)
+
+	isNew := false
+	if len(caseType.ID) == 0 {
+		isNew = true
+	}
```
ludydoo, line 115: minor
```
isNew := len(caseType.ID) == 0
```

api/pkg/apps/webapp/casetypes.go
```diff
@@ -132,60 +70,120 @@ func (s *Server) CaseType(w http.ResponseWriter, req *http.Request) {
...
+
+	if err := req.ParseForm(); errResponse(w, err) {
+		return
+	}
```
ludydoo, line 119: I think the errResponse thing is a bit hard to read..

ludydoo (changes requested):

api/pkg/apps/cms/casetype_validation.go
nilueps: I assume that means adding a subroutine in the Status method to check for uuid errors and return a different error type in that case? The front-end user can't correct that type of error, right?

Is this really necessary though? All UUIDs are generated by the same library

api/pkg/apps/cms/case_post.go
```diff
@@ -15,6 +16,14 @@ func (s *Server) PostCase(w http.ResponseWriter, req *http.Request) {
 	}
 
 	kase := &payload
+
+	errList := ValidateCase(kase, &validation.Path{})
+	if len(errList) > 0 {
+		status := errList.Status(http.StatusUnprocessableEntity, "invalid case")
+		s.JSON(w, status.Code, status)
+		return
```
ludydoo, line 24: s.Error(w, status)

api/pkg/apps/cms/casetype_validation.go
ludydoo:
```
func ValidateUUID(uuid string, path validation.Path) validation.ErrList {...}
```

ludydoo (approved):

PR #103, opened 2021-07-13
COR-112: Uganda demo preparation

shinroo:
- Hardcode actual teams and users that will be using the app
- Add party attributes
- Add case templates

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 api/certs/cert.pem
 api/certs/key.pem
 */coverage.out
+api/.idea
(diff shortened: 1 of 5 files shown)


shinroo:
@ludydoo @keth @nilueps I have added TODOs in the sections where we will need to make changes in the future, predominantly with the types of attributes. Currently we only have string attributes, which will be fine for demo purposes but it would make sense if the attributes could have the same types as the case template fields!

nilueps:
Looks good to me! 

However, saved cases aren't accessible for some reason. Seems they are saved correctly because they show up in mongo, and in the list, but I get a error when I click on one. Can you confirm?

nilueps:
OK, I think I figured out what's happening. 
In your last commit you removed a bunch of parties from the static Individuals in `seeder/static.go`, including the one I was using to log in. Seems like the auth still accepted my token even though that individual was removed and I was able to create content under her ID. Which caused an error when webapp queried mongo with it (as the case's CreatorID)

@ludydoo can you imagine this type of scenario being a security concern going forward?

nilueps (approved):

PR #107, opened 2021-07-15
COR-127: Limit party dropdown options

shinroo:
Add javascript code to only render first n elements of party dropdowns

### DIFF
--- a/api/pkg/apps/webapp/templates/individual.gohtml
+++ b/api/pkg/apps/webapp/templates/individual.gohtml
@@ -355,7 +355,7 @@
                     if (res.response.items.length > 0) {
                         let n = 0
                         res.response.items.forEach(party => {
-                          if (n < 5) {
+                          if (n < {{ .Constants.PartyDropdownLimit }}) {
                             let newListItem = document.createElement('li')
                             newListItem.classList.add('dropdown-item')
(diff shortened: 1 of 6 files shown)


api/pkg/apps/webapp/templates/individual.gohtml
```diff
@@ -349,23 +349,27 @@
                     let partySelectorList = document.getElementById("partySelectorList")
                     partySelectorList.innerHTML = ''
                     if (res.response.items.length > 0) {
+                        let n = 0
                         res.response.items.forEach(party => {
+                          if (n < 5) {
```
nilueps, line 354: I think we should avoid using "magic numbers" as much as possible, it makes the code harder to maintain.
What about creating a `Constants` struct in the webapp and passing it to the renderer? Then having something like `if (n < {{.Constants.NumPartyDropdownItems}})`

nilueps (changes requested):

api/pkg/apps/webapp/templates/individual.gohtml
shinroo: That makes sense, will make the change and re-request review!

nilueps (approved):

PR #108, opened 2021-07-15
COR-150 allow core to run without TLS

ludydoo:


shinroo (approved):
Looks good to me, I guess it might be useful to have a separate docker-compose for the https version in case we need that for local testing?

keth (approved):

PR #110, opened 2021-07-15
Cor 151 reload mongo password/username when file changes

ludydoo:


shinroo (changes requested):
Hey @ludydoo should the teams store also be updated with the new functional appraoch?

ludydoo:
@shinroo teams are parties, so it's already handled in the partystore ;)

shinroo (approved):

PR #116, opened 2021-07-16
Cor 139 cms integration tests

nilueps:
Didn't add tests for comments because I guess we are going to change that module significantly?


nilueps:
29.8%.... @ludydoo  you're going to have to show me how to abstract the logic in both server_test files XD

api/pkg/apps/cms/case_post.go
```diff
@@ -15,7 +15,9 @@ func (s *Server) PostCase(w http.ResponseWriter, req *http.Request) {
 	}
 
 	kase := &payload
-	kase.ID = uuid.NewV4().String()
+	if len(kase.ID) == 0 {
+		kase.ID = uuid.NewV4().String()
+	}
```
ludydoo, line 20:
```suggestion
	kase.ID = uuid.NewV4().String()
```
All IDs have to be generated by the server ;)

api/pkg/apps/cms/case_put.go
```diff
@@ -25,8 +25,7 @@ func (s *Server) PutCase(w http.ResponseWriter, req *http.Request) {
 	}
 
 	kase.Done = payload.Done
-	kase.ParentID = payload.ParentID
-	kase.TeamID = payload.TeamID
+	kase.FormData = payload.FormData
```
ludydoo, line 28: Nice find. Case team and parent cannot be updated

api/pkg/apps/cms/case_test.go
```diff
@@ -0,0 +1,165 @@
...
+		if err != nil {
+			s.T().FailNow()
+		}
+	}
+
+	s.Run("by party type", func() { s.testCaseFilterBy("PartyIDs")(cases, parties) })
+	s.Run("by team", func() { s.testCaseFilterBy("TeamIDs")(cases, teams) })
+	s.Run("by case type", func() { s.testCaseFilterBy("CaseTypeIDs")(cases, caseTypes) })
+	s.Run("by parent", func() { s.testCaseFilterBy("ParentID")(cases, parents) })
+	s.Run("by done", func() { s.testCaseFilterByDone(cases) })
```
ludydoo, line 108: When you run subtests, you need to pass the `testing.T` to the submethod, otherwise you'll run into some weird stuff

```
s.T().Run({NAME}, func(t *testing.T){
   ... at this point, do not use `s.T()`, but rather `t`
})
s.T().Run("by party type", func(t *testing.T) { s.testCaseFilterBy("PartyIDs")(t, cases, parties) })

```

api/pkg/apps/cms/casetype_post.go
```diff
@@ -15,7 +15,10 @@ func (s *Server) PostCaseType(w http.ResponseWriter, req *http.Request) {
 	}
 
 	caseType := &payload
-	caseType.ID = uuid.NewV4().String()
+
+	if len(caseType.ID) == 0 {
+		caseType.ID = uuid.NewV4().String()
+	}
```
ludydoo, line 21: Don't change this!

api/pkg/apps/cms/casetype_test.go
```diff
@@ -0,0 +1,100 @@
...
+	// Prepare test data
+	for i, caseType := range caseTypes {
+		n := i % len(partyTypes)
+		caseType.PartyTypeID = partyTypes[n]
+		_, err := s.client.CaseTypes().Create(s.ctx, caseType)
+		assert.NoError(s.T(), err)
+	}
+
+	s.Run("by party type", func() { s.testCaseTypeFilterByPartyType(caseTypes, partyTypes) })
+}
```
ludydoo, line 80: Use the `s.T().Run("by paty type", func(t *testing.T) ..` thing again :)

api/pkg/apps/cms/server.go
```diff
@@ -38,13 +39,18 @@ func NewServer(ctx context.Context, o *server.GenericServerOptions) (*Server, er
 		return nil, err
 	}
 
+	var hydraAdmin admin.ClientService
+	if o.HydraAdminClient != nil {
+		hydraAdmin = o.HydraAdminClient.Admin
+	}
```
ludydoo, line 45: Why whould this be nil ?

api/pkg/apps/cms/server_test.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	var mongoClientFn = func(ctx context.Context) (*mongo.Client, error) {
+		mongoClient, err := mongo.NewClient(options.Client().SetAuth(options.Credential{Username: mongoUsername, Password: mongoPassword}).SetHosts([]string{mongoHost}))
+		if !assert.NoError(s.T(), err) {
+			s.T().Fatal(err)
+		}
+		if err := mongoClient.Connect(ctx); err != nil {
+			logrus.WithError(err).Errorf("failed to connect to mongo")
+			return nil, err
```
ludydoo, line 71: You don't need to use logrus here, if you need logging during a test, you can use 
```
s.T().Log(...
```

ludydoo: Otherwise `go` doesn't know which test your logging belongs to

api/pkg/apps/cms/server_test.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	s.ctx = ctx
+	s.server = srv
+	s.client = NewClientSet(&rest.RESTConfig{
+		Scheme: "http",
+		Host:   "localhost:" + port,
+		Headers: map[string][]string{
+			"X-Authenticated-User-Subject": {"mock-auth-user"},
+		},
```
ludydoo, line 96: I think we should extract this method in a package, perhaps `pkg/testing/utils.go`?

ludydoo: Or perhaps `pkg/testutils` to not have a conflict with the go standard `testing` package

api/pkg/apps/iam/server_test.go
```diff
@@ -60,18 +61,21 @@ func (s *Suite) SetupSuite() {
...
-	if err := mongoClient.Connect(ctx); !assert.NoError(s.T(), err) {
-		s.T().Fatal(err)
-		return
+	var mongoClientFn = func(ctx context.Context) (*mongo.Client, error) {
+		mongoClient, err := mongo.NewClient(options.Client().SetAuth(options.Credential{Username: mongoUsername, Password: mongoPassword}).SetHosts([]string{mongoHost}))
+		if !assert.NoError(s.T(), err) {
+			s.T().Fatal(err)
+		}
+		if err := mongoClient.Connect(ctx); err != nil {
+			logrus.WithError(err).Errorf("failed to connect to mongo")
```
ludydoo, line 70: Same thing for logging here

ludydoo (changes requested):

ludydoo:
To abstract the test logic, you could create a 

```
type GenericServerTestSuite struct {
 *suite.Suite
}
```

In which you bootstrap mongo, etc.

Then 

```
type CmsTestSuite struct {
 *testutils.GenericServerTestSuite
}
```

api/pkg/apps/cms/case_test.go
nilueps: if I understand this code snippet from the testify docs, it's actually the other way around:
```golang
	for _, t := range []struct {
		testName string
	}{
		{"first"},
		{"second"},
	} {
		suiteT := suite.T()
		suite.Run(t.testName, func() {
			// We should get a different *testing.T for subtests, so that
			// go test recognizes them as proper subtests for output formatting
			// and running individual subtests
			subTestT := suite.T()
			suite.NotEqual(subTestT, suiteT)
		})
		suite.Equal(suiteT, suite.T())
	}
```

api/pkg/apps/cms/server.go
nilueps: No idea, I was running into this when running the tests so I added this.

api/pkg/apps/cms/case_test.go
nilueps: Tests run fine btw

ludydoo (approved):

PR #117, opened 2021-07-16
COR-128 core on backbone

ludydoo:
Had to change a few things
1. Switched to using a different sessionManager
2. Updated redigo library
3. Added hash + encryption keys (for rotation) for sessionManager
4. Added a private + public oauth2 config, so that core can talk to hydra in the cluster rather than with the public address


api/go.mod
```diff
@@ -3,20 +3,18 @@ module github.com/nrc-no/core
...
 require (
-	github.com/alexedwards/scs/redisstore v0.0.0-20210606090158-85ec2fab6bdf
-	github.com/alexedwards/scs/v2 v2.4.0
+	github.com/boj/redistore v0.0.0-20180917114910-cd5dcc76aeff
 	github.com/coreos/go-oidc/v3 v3.0.0
-	github.com/fsnotify/fsnotify v1.4.7
 	github.com/go-openapi/runtime v0.19.29 // indirect
 	github.com/golang/protobuf v1.5.2 // indirect
-	github.com/gomodule/redigo v1.8.5
+	github.com/gomodule/redigo v2.0.0+incompatible
```
nilueps, line 10: is this leftover from when you thought redigo was not working?

nilueps (approved):

PR #120, opened 2021-07-21
COR-160: Taxonomy picklists

shinroo:
- add case element type `taxonomyinput`
- configure taxonomy input for Uganda use-case

Demo:
Peek 2021-07-21 16-41


api/pkg/apps/webapp/templates/form_element.gohtml
```diff
@@ -60,5 +60,261 @@
...
+            const clearSelectOptions = (selectId) => {
+              document.getElementById(selectId).innerHTML = "";
+            }
+
+            // create a new select option element
+            const createSelectOption = (value, label) => {
+              const option = document.createElement("option");
+              option.value = value;
+              option.text = label;
+              return option
```
keth, line 166: There are missing `;` after return statements in this file.  
For long .js snipets like this can we put it in a sepperate file?  something like this >>
`<script src="./js/myScript.js" type="text/javascript" charset="utf-8"></script> `

nilueps: I agree, we should start putting scripts like this one in their own folder 👍🏻

nilueps (changes requested):
Looking good! Could you add a check to prevent the user from adding a duplicate services?

api/pkg/apps/webapp/templates/form_element.gohtml
shinroo: That makes sense! I will implement this

shinroo:
> Looking good! Could you add a check to prevent the user from adding duplicate services?

Yep for sure! I think I will also add the ability to delete a single item by clicking on it? Perhaps have it turn red on hover to show that it can be deleted.

api/pkg/apps/webapp/templates/form_element.gohtml
shinroo: The only downside is that we can then not use golang templatign directives in the JS code.

shinroo (commented):
Demo:
Peek 2021-07-22 13-02

api/pkg/apps/webapp/templates/form_element.gohtml
nilueps: Good point. So instead of importing a .js script, why not template the entire `<script>` block?

nilueps (approved):

keth (approved):

PR #125, opened 2021-07-27
COR-113: User cannot login after setting credential

shinroo:
Diagnosed problem where user could not login after setting credentials:
- Null request was being sent to Hydra from app
- Null request happened because we did not store credential in MongoDB

To remedy:
1. Added a new method called PostCredential() which takes care of hashing, salting and storing the password in MongoDB

Peek 2021-07-27 13-26

### DIFF
--- a/api/pkg/apps/login/credentials_post.go
+++ b/api/pkg/apps/login/credentials_post.go
@@ -1,0 +1,26 @@
+package login
+
+import (
+	"net/http"
+)
+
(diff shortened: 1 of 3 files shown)


shinroo (commented):
@nilueps I have now handled the situation you discovered yesterday and I have also cleaned up the PR a bit

nilueps (approved):

PR #126, opened 2021-07-29
Cor 126 render validation

nilueps:
Ok, so I thought this would be easier but it turned into kind of a little monster. You'll probably want to clone this one rather than just review the diffs herer on GH.

The problem this PR is supposed to solve is simply to handle render form validation on the front end by applying bootstrap classes and displaying error messages to the User.

In solving this problem I refactored a couple functions, mainly the `PostCase` handler. I also refactored the form field templates and introduced a `Readonly` boolean field on the `FormElement` struct of the `CaseTemplate`s which, for the moment, is applied to a `Case` when it is retrieved from the DB. Since we are using the same template to display new case forms as well as filled-out case forms, the presence of this flag makes it easy to distinguish the two on the front end.


shinroo (commented):
Wow this is a beast of a PR - I am having a bit of trouble getting it to run in my Goland. When I switch to this branch all the run configurations seem to disappear.

shinroo (approved):
Got Goland to work with it, but had to restart.

The validation looks good to me!

Peek 2021-07-30 13-03

PR #127, opened 2021-07-29
Cor 174 registration controller

nilueps:
This PR introduces the `registrationcontroller` package. This package exposes a limited interface in an attempt to limit its dependencies.

The main object is `RegistrationController`. Its constructor expects a `CaseHandler` and a `CaseFlow`. The former describes an interface for retrieving case information (state) which should be implemented by the caller; the latter is an ordered list of `CaseType` that describes the expected case steps the controller should use to interpret the state.

The information provided by `RegistrationController` is defined in the `Action` and `Status` types. That information is rather bare for now, let me know if you think of something I should add.

The skeleton for tests is there, but I wanted to make sure I was on the right track before inverting too much time testing something that doesn't correspond to the requirements.


shinroo (changes requested):
Hey Nico, this all looks good so far. 

I left a comment on the JIRA ticket yesterday but I will rephrase it here:

For Uganda the intake procedure will use Attachments and not Cases.

The three steps for the Uganda intake are:
1. Fill out individual attributes
2. Add situational analysis as attachment to individual and fill out
3. Add individual assessment as attachment to individual and fill out

We should sync up on how we can link up our two workstreams to enable this!

shinroo (dismissed):
Nico and I had a sync, and we agreed to keep the controller as is for now and pause the usage of the attachments API as it might make more sense to refactor out the templates into a top level form entity which could then be shared by Cases and attachments.

We believe this is beneficial, since the attachments used in the registration process are very similar/identical in function to the case templates and we would need to duplicate a lot of work if we created similar logic within the attachments. Therefore we think splitting off the templates makes sense.

For now, these changes are accepted so that we can make progress on UI customisations for Uganda!

api/.idea/workspace.xml
```diff
@@ -1,36 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
shinroo, line 1: What's up with this file?

shinroo (approved):

api/.idea/workspace.xml
nilueps: Not sure where it came from, but we don't need nested intellij configs afaik

PR #128, opened 2021-07-30
COR-147: Attachment API

shinroo:
- add new app `attachments`
- add integration tests, including list filters

**NB:** This feature will be further developed in future, and this is intended as a starting point!

Test results:
```
/usr/local/go/bin/go tool test2json -t /tmp/___1test_attachments_integration.test -test.v -test.paniconexit0
=== RUN   TestSuite
--- PASS: TestSuite (3.16s)
=== RUN   TestSuite/TestAttachments
    --- PASS: TestSuite/TestAttachments (2.70s)
=== RUN   TestSuite/TestAttachments/API
        --- PASS: TestSuite/TestAttachments/API (0.33s)
=== RUN   TestSuite/TestAttachments/List_filtering
        --- PASS: TestSuite/TestAttachments/List_filtering (2.21s)
PASS

Process finished with the exit code 0
```


api/pkg/apps/attachments/interface.go
```diff
@@ -0,0 +1,22 @@
...
+func (a *AttachmentListOptions) MarshalQueryParameters() (url.Values, error) {
+	ret := url.Values{}
+	ret.Set("attachedToId", a.AttachedToID)
+	return ret, nil
+}
+
+func (a *AttachmentListOptions) UnmarshalQueryParameters(values url.Values) error {
+	a.AttachedToID = values.Get("attachedToId")
+	return nil
+}
```
nilueps, line 22: What's the reason for placing these in this file? I don't think they are part of the package's interface. If anything they should be private methods (lowercase first letter)

nilueps (changes requested):
Just the one  comment, otherwise looks good to me!

api/pkg/apps/attachments/interface.go
shinroo: I was inspired by this file: https://github.com/nrc-no/core/blob/74d0e2f6c2080d5fdfa656bf5e6702992a9135b9/api/pkg/apps/iam/client_interface.go

Should I change them to being private?

shinroo: I could also move the Marshal/Unmarshal to a separate file, what do you suggest?

nilueps: Yeah, there are lots of methods/functions all over the place which I think should be private. If they are private they could stay I think. Otherwise maybe they should go somewhere else. Other option: rename the file :P ..after all the package's interface goes beyond what's in this file

nilueps (approved):

PR #130, opened 2021-08-02
COR-177: add logic to prevent dropdown list getting hidden on click

shinroo:
The party selector list now no longer gets hidden when clicking on the focused input.

Please excuse the google autofill dropdown that gets rendered by Chrome over the dropdown in the demo:
Peek 2021-08-02 13-46

### DIFF
--- a/api/pkg/apps/webapp/templates/head.gohtml
+++ b/api/pkg/apps/webapp/templates/head.gohtml
@@ -6,13 +6,13 @@
         <meta name="viewport" content="width=device-width, initial-scale=1">
 
         <!-- Bootstrap CSS -->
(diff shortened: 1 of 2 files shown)


nilueps (changes requested):
Looks like the dropdown doesn't disappear, but the content is not selectable!
Also notice a bug for another ticket: nothing prevent the user from creating a relationship to a non-existent party (see gif below)

Peek 2021-08-02 08-57

nilueps (approved):

nilueps:
This took way longer than it should have 😅 
I approved it, but you may way to double-check it @shinroo

PR #131, opened 2021-08-02
COR-173: Country API

senyaoh:
- Country API
- Display corresponding country in team page


.idea/runConfigurations/iam_coverage_report.xml
```diff
@@ -7,7 +7,7 @@
     <option name="INDEPENDENT_SCRIPT_WORKING_DIRECTORY" value="true" />
     <option name="SCRIPT_WORKING_DIRECTORY" value="$PROJECT_DIR$/api/pkg/apps/iam" />
     <option name="INDEPENDENT_INTERPRETER_PATH" value="true" />
-    <option name="INTERPRETER_PATH" value="/usr/bin/bash" />
+    <option name="INTERPRETER_PATH" value="/usr/local/bin/bash" />
```
shinroo, line 10: We keep getting changes to these files

api/pkg/apps/webapp/teams.go
```diff
@@ -65,6 +65,20 @@ func (s *Server) Team(w http.ResponseWriter, req *http.Request) {
...
 
+	n, err := iamClient.Nationalities().List(ctx, iam.NationalityListOptions{
+		TeamID: id,
+	})
+	if err != nil {
+		s.Error(w, err)
+		return
+	}
+
+	c, err := iamClient.Countries().Get(ctx, n.Items[0].CountryID)
```
shinroo, line 76: Can we add a check here on the n.Items to make sure that its len is bigger than 0 before we address n.Items[0], this may lead to a panic like we saw earlier

shinroo (changes requested):
Looks good to me, a few minor things to take a look at.

.idea/runConfigurations/iam_coverage_report.xml
senyaoh: I'm not sure what to do about this one, it depends on the OS and I guess we use different ones...

shinroo: Nothing to do, but we need to fix it at a higher level

shinroo (approved):

PR #133, opened 2021-08-03
COR-167: UI Modifications for UAT in UG

shinroo:
This PR is a bulk one for final preparations before we begin UAT in Uganda.

It mainly intends to move a subset of cases, namely those involved in the intake process, into the individual registration flow.

This is achieved by using work completed previously including the Registration Controller. The UI of the individual view is modified to include the cases beneath the Attributes.

It also fixes some issues:
- re-adding taxonomyinput case element type

This PR was worked on in collaboration with @nilueps

### DIFF
--- a/.idea/core.iml
+++ b/.idea/core.iml
@@ -20,5 +20,8 @@
     </content>
     <orderEntry type="inheritedJdk" />
     <orderEntry type="sourceFolder" forTests="false" />
+    <orderEntry type="library" name="bootstrap" level="application" />
+    <orderEntry type="library" name="bootstrap-icons" level="application" />
+    <orderEntry type="library" name="rxjs" level="application" />
   </component>
 </module>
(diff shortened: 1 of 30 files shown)


nilueps:
Ok, so I think this _could_ be merged as is, at least insofar as we're able to create new individuals who get preassigned blank intake cases.
There's still no validation on attributes, and we need to figure out how to get the attribute section's status to go from `Incomplete` to `Closed` (it currently remains as  `Unopened`)

shinroo:
> Ok, so I think this _could_ be merged as is, at least insofar as we're able to create new individuals who get preassigned blank intake cases.
> There's still no validation on attributes, and we need to figure out how to get the attribute section's status to go from `Incomplete` to `Closed` (it currently remains as `Unopened`)

On the validation for attributes, lets tackle that with the form refactor in a separate ticket.

As for the status, perhaps we should make a ticket for that as well. Since it probably relates to the attributes validation working as expected.

I think for UAT the current status should be fine!

shinroo (commented):
Found an issue where the long formulation is not used when the value is not set, fixing in a commit shortly...

Screenshot from 2021-08-06 00-12-07

shinroo (commented):
Demo of latest version as of yet:
Peek 2021-08-06 00-15

PR #138, opened 2021-08-09
COR-149: add registration status label to invididual list page

shinroo:
### Original ticket description:

As a { Project Manager | Enumerator | Triage Staff | Expert }
I want to see a beneficiaries status from the beneficiary list page
So that I can take pertinent actions for this beneficiary

Technical Description:

Beneficiary entry should show status. If status is an action, status should be a button pointing to form.

### Note

Since we did not yet go down the route of using actions, the later part of the requirement:

> If status is an action, status should be a button pointing to form.

Has not been implemented.

### DIFF
--- a/api/pkg/apps/webapp/individuals.go
+++ b/api/pkg/apps/webapp/individuals.go
@@ -7,6 +7,7 @@ import (
 	"github.com/nrc-no/core/pkg/apps/cms"
 	"github.com/nrc-no/core/pkg/apps/iam"
 	"github.com/nrc-no/core/pkg/apps/seeder"
+	"github.com/nrc-no/core/pkg/registrationctrl"
 	"github.com/nrc-no/core/pkg/sessionmanager"
 	uuid "github.com/satori/go.uuid"
 	"golang.org/x/sync/errgroup"
@@ -14,6 +15,11 @@ import (
 	"strconv"
 	"strings"
 )
+
+type IndividualWithStatuses struct {
+	Individual         *iam.Individual
+	RegistrationStatus *registrationctrl.Status
+}
 
 func (s *Server) Individuals(w http.ResponseWriter, req *http.Request) {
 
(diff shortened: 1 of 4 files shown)


shinroo:
Screenshot from 2021-08-09 15-45-39

api/pkg/apps/webapp/individuals.go
```diff
@@ -48,9 +54,23 @@ func (s *Server) Individuals(w http.ResponseWriter, req *http.Request) {
...
+			RegistrationStatus: rc.Status(),
+		})
+	}
+
 	if err := s.renderFactory.New(req, w).ExecuteTemplate(w, "individuals", map[string]interface{}{
-		"Individuals": list,
-		"Page":        "list",
+		"Individuals":             list,
+		"IndividualsWithStatuses": beneficiaryWithStatuses,
+		"Page":                    "list",
```
nilueps, line 73: Is it necessary to pass in an entirely new pipe?

```
type Individual struct {
    *iam.Individual
    Status *registrationctrl.Status
}
```

Then

```
{{range .Individuals}}
    ...
    {{if .Status}}...{{end}}
{{end}}
```

nilueps (commented):
This works, but I think you can simplify the handler a little, see comment

api/pkg/apps/webapp/individuals.go
shinroo: Currently, `Individuals` is an individual list.

The pagination relies on that in the following code segment:
```
{{template "pagination-navigation" .Individuals.Metadata }}
```

I had initially tried to just pass a map of `id` -> `registrationctrl.Status` and then simply call that from the individual row. But I wasn't able to get that working.

nilueps: Ah right right, well in that case I think your solution is fine

nilueps (approved):

PR #139, opened 2021-08-10
Cor 190 fix tests

nilueps:
- Fixed iam and cms tests
- Rewrote Attribute, CaseType, Case and Individual e2e tests in large part to add verification tests to ensure entered data was properly being saved and rendered.
- Reported encountered bugs (201-202)
- Integration tests are ✅ but e2e suite is ❌ due to frontend bug described in 201


senyaoh:
@nilueps All the integration tests ran smoothly without problems. As for the e2e tests, 7 out 27 failed. The failed ones are for the attribute page and the individual page.
Screen Shot 2021-08-12 at 00 42 33

senyaoh:
@nilueps Any reason the `TestCountry` and `TestNationality` are singled out options in goland for testing? Could we remove them>

PR #146, opened 2021-08-13
Fixup broken case submission

nilueps:
The structure of cases.go was/is a mess. Hopefully the logic is a bit cleaner now. The behaviour of the api is also more obvious, with POST requests returning JSON validation objects when errors are encountered. The frontend still needs to be updated to handle this data appropriately in the way that the `individual.gohtml` template currently does. As an aside, I believe it might be worthwhile to move some of the validation logic to the client side, at least as far as syntax checking is concerned (emails, phone numbers, etc). Given that Core is intended to function in environments with spotty connection, we should reserve server-side validation for evaluations which are impossible to do on the client host.


api/pkg/apps/webapp/cases.go
```diff
@@ -94,151 +94,63 @@ func (s *Server) Cases(w http.ResponseWriter, req *http.Request) {
...
-		team = teamRes
-	}
-
-	// Get parent case
-	if len(kase.ParentID) > 0 {
-		parent, err = cmsClient.Cases().Get(ctx, kase.ParentID)
-		if err != nil {
+	if !ok || len(caseID) == 0 {
+		if req.Method == "POST" {
+			caseID = "new"
```
keth, line 121: we probably want to use Constant for "new"

api/pkg/apps/webapp/cases.go
```diff
@@ -247,8 +159,112 @@ func (s *Server) Case(w http.ResponseWriter, req *http.Request) {
 				s.Error(w, err)
 			}
 			return
+		}
+		var action string
+		if caseID == "new" {
+			action = "created"
```
keth, line 165: Missing Constants for all action types

api/pkg/apps/webapp/cases.go
```diff
@@ -363,39 +377,35 @@ func (s *Server) NewCase(w http.ResponseWriter, req *http.Request) {
 	}
 }
 
-func (s *Server) PutOrPostCase(req *http.Request, w http.ResponseWriter, ctx context.Context, kase *cms.Case) error {
+func (s *Server) PostCase(req *http.Request, w http.ResponseWriter, ctx context.Context, kase *cms.Case) (*cms.Case, error) {
```
keth, line 380: case with "k" ??  is this typo?

api/pkg/apps/webapp/cases.go
```diff
@@ -264,9 +280,7 @@ func (s *Server) Case(w http.ResponseWriter, req *http.Request) {
 		"Comments":         displayComments(comments, commentAuthorMap),
 	}); err != nil {
 		s.Error(w, err)
-		return
 	}
-
 }
```
keth, line 284: This method is more than 150 lines of code.  I think we can chop it up a little, let's take a pair programming session together.

api/pkg/apps/webapp/cases.go
nilueps: hehehe I guess you might also have been pulling your hair out trying to figure out why you were getting really weird syntax errors until you remember that `case` is a keyword :P

api/pkg/sessionmanager/interface.go
```diff
@@ -85,33 +85,37 @@ const (
...
 func (r RedisSessionManager) AddNotification(req *http.Request, w http.ResponseWriter, notification *Notification) error {
-	session, err := r.Store.Get(req, varNotifications)
-	if err != nil {
-		return err
-	}
-	session.AddFlash(notification)
-	if err := session.Save(req, w); err != nil {
-		return err
-	}
+	// FIXME was converted to noop because of securecookie error
```
shinroo, line 88: Have we created tickets for these FIXMEs in JIRA?

nilueps: Now we have :)

keth (approved):

PR #147, opened 2021-08-16
Cor 62 local test hooks

nilueps:
To install hook run the `install.sh` script in the `actions` folder. This will sym-link the `pre-push` hook to `.git/hooks` and the hook will run before each push.

For now, the hook runs the `iam` and `cms` integration tests only because the e2e tests are not ready. If a test fails, git will not let you push your code.

This PR includes some updates to e2e tests as well as disabling certain tests until they are updated to test the upcoming UI changes


.idea/jsLinters/eslint.xml
```diff
@@ -2,5 +2,6 @@
 <project version="4">
   <component name="EslintConfiguration">
     <option name="fix-on-save" value="true" />
+    <files-pattern value="{**/*,*}.{js,ts,jsx,tsx,html,vue,gohtml}" />
```
shinroo, line 5: Vue?

actions/git-hooks/pre-push
```diff
@@ -0,0 +1,99 @@
...
+	netstat -tln | grep -q ":9000" &> /dev/null
+	return $?
+}
+if ! serverIsRunning; then
+	echo "$PREFIX webapp is not running, attempting to start it now."
+	cd "$DIR/api" || exit
+	make serve &> /dev/null &
+
+	n=0
+  until [ "$n" -ge 5 ]
```
shinroo, line 67: Seems like the indentation got a bit messed up here

actions/git-hooks/pre-push
```diff
@@ -0,0 +1,99 @@
...
+
+# linting
+#
+echo "$PREFIX Running eslint on .js files..."
+if ! npm run lint:fix; then
+	>&2 echo "[pre-push] Linting failed, see output above."
+	exit 1
+fi
+
+echo "$PREFIX Running E2E tests..."
```
shinroo, line 91: maybe we can skip the e2e until the new design comes in and we fix the broken tests?

shinroo (changes requested):
We should rename `party_test.go.bk` and `partytype_test.go.bk` back to their original names so they are included in the test suite.

.idea/jsLinters/eslint.xml
nilueps: This was auto-generated after I changed some settings, not sure which tbh, but I think it's harmless

actions/git-hooks/pre-push
nilueps: I opted to skip the invalid tests themselves, that way we at least have coverage for some of the pages

shinroo (approved):

PR #151, opened 2021-08-18
Cor 186 Prevent intake case editing until attribute are filled out

nilueps:
This turned out to be kind of redundant after COR-187 Registration Subpages because the intake case tab buttons remain disabled until the attributes stage has been completed. Nevertheless, the webapp's behavior now further ensures the Cases passed to the renderer are marked as readonly if the attributes haven't been filled out.


shinroo (approved):
Looks good to me!

Peek 2021-08-19 12-28

PR #157, opened 2021-08-25
COR-224: Added new intake attributes

senyaoh:
Added following individual intake attributes:
- Date of Identification (Date)
- Location of Identification (dropdown)
- Source of Identification (dropdown)
- District / Admin 2 (dropdown)
- Subcounty / Admin 3 (text)
- Parish / Admin 4 (text)
- Village / Admin 5 (text)

### DIFF
--- a/api/pkg/apps/seeder/static.go
+++ b/api/pkg/apps/seeder/static.go
@@ -491,6 +491,74 @@ var (
 					},
 				},
 			},
+			{
+				Type: form.Date,
+				Attributes: form.FormElementAttributes{
+					Label:       "Date of Identification",
+					Name:        "identificationDate",
+					Description: "Date of first interaction with NRC",
+				},
+			},
+			{
+				Type: form.Dropdown,
+				Attributes: form.FormElementAttributes{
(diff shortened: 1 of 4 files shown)


api/pkg/apps/seeder/static.go
```diff
@@ -491,6 +491,74 @@ var (
 					},
 				},
 			},
+			{
```
shinroo, line 494: These changes should not be made to the `UGICLAIndividualIntake` case template, but should be added as attributes.

An example of the global attributes can be seen here: https://github.com/nrc-no/core/blob/2f03a416ed2f2bc90837e7fa748773a74bdf518d/api/pkg/apps/seeder/static.go#L76

Further, this ticket needs to introduce a mechanism to handle country-attributes. The existing attributes are all global (apply to all countries). The attributes in this ticket are specific to the Ugandan context.

The user should be presented with the combination of the global attributes and the country attributes (inferred from the country/countries they are associated with via the team memberships they have).

shinroo (changes requested):

shinroo (approved):

PR #158, opened 2021-08-27
Cor 206 handle case attr validation

nilueps:
# Case and Attribute validation handling

This ticket's goal was to modify the Case and Attribute front ends to consume server-side validation objects when received. We gave the ticket a score of **2**, but it probably deserved an **8** or more as it turned out to be a somewhat complex task.

For starters, I wanted to extract the front-end validation handling logic which was located in the `individual.gohtml` template into a reusable JS library we could call from any template. I could have created a dedicated template for this, but I opted to write the lib in a `.ts` file to benefit from syntax highlighting and type checking. To achieve this I added a method on the `webapp` server to transpile `.ts` files on startup and added an endpoint from which to serve javascript files: `/static/js/{filename}`.

The `validation.ts` mini-library exports two functions `validateClientSide` and `validateServerSide`. I modified the `individual`, `case`, `casenew` and `attribute` templates to make use of these functions.

The idea is that we make an array of `HTMLFormElement`s on which we first perform client-side validation relying on the [Constrain Validation API](https://developer.mozilla.org/en-US/docs/Web/API/Constraint_validation) to catch errors and [Bootstrap Validation](https://getbootstrap.com/docs/5.0/forms/validation/) to display them.

If `validateClientSide` returns `true` -- that is, if no errors are encountered -- we proceed to collect form data, including data from custom form elements which we currently cannot validate on the client, and perform POST requests for each form. If the server responds with validation errors (JSON format), we update the UI accordingly. If no validation is returned, we proceed to redirect the browser.

### Attribute validation

You will notice I commented out an attribute validation step as well as the tests associated with that behaviour.

When working on the attributes, I realized we were calling the same `ValidateAttribute` server-side method in two situations which actually should receive different types of validation. The first situation is when submitting an entirely new Attribute via the Attributes API (e.g. a new attribute called "isEmployed" with translation info and other caracteristics). The second is when submitting actual attribute data when submitting an Individual's attributes (e.g. setting "isEmployed" to "true" on John Doe). In both cases, we were checking whether the submitted attribute was required and, if so, whether it had a value.

Obviously, newly created attributes will not have a value, but if they are marked as required, they will fail validation.

Because server-side attribute value validation involves checking the given value against the attribute's characteristics, it makes sense to involve the `Attribute` struct in that validation process.  However, I would propose we implement this validation elsewhere, not in the attribute module itself.

Until then, I simply disabled that step which doesn't impact UX since client-side validation catches that kind of error before any POST. And opened a ticket to address this problem.

### Individual e2e test

As me and @shinroo noted recently. There is some weirdness happening with the individual e2e test. The test, which essentially goes through the steps of registering a new beneficiary, behaves unpredictably; sometimes passing, sometimes failing. When the test fails, it is either because it created the individual but the relationship it added to it was not saved,  or it is because two duplicate individuals were created, one without the relationship and open cases, the other with no relationship and closed, empty, cases.

Me and @shinroo spent quite some time trying to narrow down the problem, as did I again today. I'm still not sure what's going, but it's most probably something to do with cypress itself and not our test.

I instructed cypress to skip the individual test so that I could push my code and opened a new ticket to find a solution.

## n.b.

If you don't have typescript installed, the server won't start. In that case this should be enough:
```bash
npm install -g typescript
```


api/pkg/apps/iam/attribute_validation.go
```diff
@@ -74,10 +74,17 @@ func (a *Attribute) ValidateFormField(errList validation.ErrorList, path *valida
...
+	// attribute) and when submitting an attribute value via API. The absence of a value in the first case was
+	// triggering an error here and there was no easy way to avoid it.
+	// It might be ok to skip this step because client-side validation prevents submitting empty required values via the
+	// front end, however, this obviously leaves the door open to direct API submissions.
+	// Ideally, editing available Attributes, and validating user-submitted Attribute _data_ should be handled in
+	// separate places.
+	//if a.Validation.Required && utils.AllEmpty(a.Attributes.Value) {
+	//	err := validation.Required(fieldPath, fmt.Sprintf("%s is required", a.Attributes.Name))
+	//	errList = append(errList, err)
+	//}
```
ludydoo, line 87: I would not remove this validation... Let's talk together and try to figure a way around this

api/pkg/apps/webapp/server.go
```diff
@@ -99,6 +100,12 @@ func NewServer(options *ServerOptions) (*Server, error) {
 		return nil, err
 	}
 
+	err = transpileTS()
+	if err != nil {
+		logrus.WithError(err).Errorf("failed to transpile typescript files")
+		return nil, err
+	}
+
```
ludydoo, line 108: I think that transpiling typescript files should rather be part of the build process..

ludydoo (commented):
I'd like to do a code review with you if possible

nilueps:
About three weeks ago, me and @ludydoo had an extensive pair programming session on this branch. What began as a code review quickly led to some extensive rewrites and refactors -- most notably in the `form` types -- which had some far-reaching consequences throughout the code-base. Hence the massive size of this PR.

These changes totally outgrew the initial purpose of this ticket which was only concerned with Case and Attribute validation.

## Form structs

The structure of the `Control` struct which represents a form control element was flattened somewhat and redundant fields have been removed. This struct's purpose is now clearer: it describes an HTML form control element, its attributes and values. This in turn allowed us to simplify the representation of `Attributes`

## Attribute & Form Control Internationalization

Another big change was the removal of the `Translations` field from the `Party` `Attribute` struct (now renamed to the less ambiguous `PartyAttributeDefinition`). Instead of storing translations in their own struct, me and @ludydoo felt it made more sense to change the types of user-facing fields from string primitives to dedicated i18n structs. For instance, a form control's `Label` is no longer described by a simple `string`, but by the `i18n.Strings` struct which represents a list of translations.

```golang
package i18n

type LocaleString struct {
	Locale string `json:"locale" bson:"locale"`
	Value  string `json:"value" bson:"value"`
}

type Strings []LocaleString
```

```golang
type Control struct {
	/* ... */
	Label           i18n.Strings      `json:"label" bson:"label"`
	/* ... */
}
```

I have updated the static assets with the new data types, but stopped short of updating the web app to make full use of these. For now, core has been hard-coded to use the English, "en" locale. Given the already monstrous dimensions of this PR, I think the proper implementation of i18n features should happen in separate tickets.

shinroo (approved):
Integration and e2e tests look good locally.

PR #159, opened 2021-08-27
COR-213 Name fields

nilueps:
I repurposed the `FirstName` and `LastName` attributes and renamed them to `FullName` and `Display Name` and updated the rest of the app accordingly. Search functionality still works


shinroo (approved):
Tested it manually locally, looks good to me!

PR #164, opened 2021-08-31
create data flow diagram

ludydoo:


shinroo (approved):
This really clarifies the flow, especially regarding the login procedure and the role of the attachments API.

PR #166, opened 2021-09-02
COR-223: Identification documents

shinroo:
- Add identification document type and indentification document as entities
- Implement APIs
- Add integration tests
- WIP implementation on the webapp integration (need some time with @nilueps)


shinroo:
Still to-do:
- be able to create ID documents from webapp
- be able to delete ID documents from webapp
- update e2e tests

internetti (changes requested):
submitting with empty document number causes 500 error

internetti (approved):

PR #167, opened 2021-09-06
COR 229 frontend input sanitization

nilueps:
This PR includes a merge of the as-yet-unapproved PR COR-206.
There's only so much we can do as far as client-side input sanitizing, the main work is going to have to happen in the backend.
As such, the only front-end sanitizing you'll find are `maxlength` attributes on the text-based inputs. Our client-side validation function already takes care of handling malformed values (dates, emails, phone, etc).

- setup ts stack
- make required textarea for quick dev testing purposes (also should be required anyway)
- update validation logic
- update ts target and libs; disable sourcemap as we don't serve ts
- fix case overwrite with nil when err
- update template to use new validation.js module
- ignore tmp folder
- add simple client-side validation
- add validation handling to template
- update gitignore
- add CustomDiv type as catchall for custom form elements
- simplify client-side validation function to make use of the Constraint Validation API
- update use of client-side validation function
- always redirect to redirectPath on successful SS validation; change validateClientSide signature to accept array of HTMLFormElement
- make use of updated validation functions; reformat ids; remove unused code
- remove conflicting validation for empty required attributes: left verbose FIXME comment in-line.
- fix validation error handling and hard-code return object structure.
- pass Status pipe variable to individual template
- comment out require field test cases
- update individual tests
- update submit event handler signature
- fix typo
- add input constraints
- Changes by anon
- massive refactor
- clean up validation; cases attributes
- main functionalities restored
- update e2e
- update e2e
- revert pre-push hook
- input sanitizing


shinroo (changes requested):
A couple thoughts on this PR:
- Would it make sense to define these limits somewhere centrally? Perhaps in a go file where we could then template these in so its like `maxlength="{{.URLMaxLength}}"`. Not high prio on this point since we will likely rewrite the frontend anyway, but might make sense :man_shrugging: 
- Github is showing me that we have a 0 length file in the PR which is not being deleted. Perhaps we need to delete it manually?

Screenshot from 2021-09-28 16-20-27

nilueps:
> A couple thoughts on this PR:
> 
>     * Would it make sense to define these limits somewhere centrally? Perhaps in a go file where we could then template these in so its like `maxlength="{{.URLMaxLength}}"`. Not high prio on this point since we will likely rewrite the frontend anyway, but might make sense man_shrugging
> 
>     * Github is showing me that we have a 0 length file in the PR which is not being deleted. Perhaps we need to delete it manually?
> 
> 
> Screenshot from 2021-09-28 16-20-27

I agree the first suggestion is a better solution, but as you point out yourself, we are going to rewrite the front end anyway, so I think we could hold off on this.

As for the second point, I'm not sure how that ended up there, but the empty file isn't part of the branch afaict, at least not on my local version. But does exist on main. I'm going to go ahead and delete it there.

shinroo (approved):

PR #168, opened 2021-09-12
COR-235: Country specific attributes

senyaoh:
- Added `CountryID` to `Attribute`. Each attribute is now given either `GlobalCountry.ID` or a specific country ID defined in `iam.static.go`
- Added `CountryIDs` to `AttributeListOptions`. By passing `CountryIDs: []string{iam.GlobalCountry.ID, countryID}` we get attributes that are either global or match countryID
- Added `GetCountryFromLoginUser` in `profile.go`, which returns the country ID of the team of the login user.
- Attributes in `Attribute Settings` and `Individuals` are now only showing global attributes and attributes specific to login user's country.

### DIFF
--- a/api/pkg/apps/iam/attribute_store.go
+++ b/api/pkg/apps/iam/attribute_store.go
@@ -3,7 +3,6 @@ package iam
 import (
 	"context"
 	"github.com/nrc-no/core/pkg/utils"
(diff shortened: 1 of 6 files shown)


shinroo (approved):
Tested locally and it looks good to me!

PR #176, opened 2021-09-29
COR-269: UG protection incidence

internetti:


api/pkg/apps/seeder/static.go
```diff
@@ -517,12 +517,88 @@ var (
...
+					},
+				},
+			},
+			{
+				Name:  "comment",
+				Type:  form.Textarea,
+				Label: i18n.Strings{{"en", "Comment"}},
+			},
+			{
+				Name:    "incidentHasBeenReportedToOthers",
```
internetti, line 583: I took the liberty here to leave out the "No" option and make the field not required.

ludydoo (approved):

shinroo (approved):

PR #188, opened 2021-09-29
COR 262 form hints

nilueps:
this doesn't fulfill the ticket requirements because of time constraints.

It doesn't add the ability to show hint dynamically, it only adds a `Hint` form control type with static text content.


shinroo (approved):
Runs fine on my end!

PR #194, opened 2021-10-01
COR-281: Remove other services and comments

senyaoh:
Remove “other services requested” in the UG response form as well as comments


shinroo (approved):
Looks good to me, works on local testing!

PR #198, opened 2021-10-06
COR 287 env vars for frontend

sqwxl:
You should now create a .env file in `core/frontend` with the local IP of the device you have the server running on

```
SERVER_HOSTNAME='xxx.xxx.xxx.xxx:9000'
```


.gitignore
```diff
@@ -7,4 +7,5 @@
 *.out
 actions/__tmp
 vendor/
-actions/__tmp
\ No newline at end of file
+api/tmp/
+api/models.ts*
```
internetti, line 11: could we also add the model backups here?

frontend/app.config.js
```diff
@@ -0,0 +1,8 @@
+import 'dotenv/config';
+
+export default {
+    extra: {
+        server_default_hostname: 'localhost:9000',
+        server_hostname: process.env.SERVER_HOSTNAME,
+    },
+};
```
internetti, line 8: how much more work would it be to enable dev/prod/local environments?

.gitignore
nilueps: that's what the _*_ is for ;)

internetti: shouldn't those files be removed from this pr then?

PR #210, opened 2021-10-14
COR-304 Floating save btn

sqwxl:
Peek 2021-10-14 14-45


nilueps:
@ludydoo wouldn't it also appear over the text in that case?

PR #216, opened 2021-10-18
COR-244: JavaScript API Client

shinroo:
Introduces a typescript client for the Core API
- typescript types are automatically generated from our Golang type definitions
- the clients use RxJS and especially `OperatorFunctions`
- all clients are unit tested to ensure that requests are properly built


nilueps:
All tests passing without issue here! :tada:

PR #221, opened 2021-10-20
COR-297 Attribute aliases

sqwxl:


nilueps:
now they look like this:

Screenshot from 2021-10-21 16-55-33

PR #234, opened 2021-11-01
COR-335 MultiSelect Field Type

nilueps:
note: removed the `bg-dark`, `text-white` classes because they weren't playing nice with the react-select component


web/pwa/src/features/admin/Organization.tsx
```diff
@@ -0,0 +1,60 @@
+import React, {FC, Fragment, useEffect} from "react";
```
ludydoo, line 1: This file should be deleted...

ludydoo (changes requested):
We should work together on this

web/pwa/src/features/recorder/FieldEditor.tsx
```diff
@@ -103,19 +104,40 @@ export const SingleSelectFieldEditor: FC<FieldEditorProps> = props => {
     </div>
 }
 
+export const MultiSelectFieldEditor: FC<FieldEditorProps> = props => {
+    const {field, value, setValue} = props
+
+    console.log("value: ", value)
```
internetti, line 110:
```suggestion
```

PR #235, opened 2021-11-02
COR-332: add month field type

shinroo:
- Adds a new field type "Month", which is a combination of a year and a month in the `YYYY-MM` format (this uniquely identifies a month)
- Validation with a `pattern` attribute on the input


web/pwa/package.json
```diff
@@ -17,6 +17,7 @@
     "axios": "^0.23.0",
     "bootstrap": "^5.1.3",
     "bootstrap-icons": "^1.6.1",
+    "inputmask": "^5.0.6",
```
ludydoo, line 20: Is this still needed ?

ludydoo (changes requested):

web/pwa/src/features/recorder/FieldEditor.tsx
```diff
@@ -71,6 +71,43 @@ export const DateFieldEditor: FC<FieldEditorProps> = props => {
...
+
+    const [localValue, setLocalValue] = useState(value != null ? (value as Date).toISOString().slice(0, 7) : "")
+
+    const isValidLength = () => localValue.length === expectedLength;
+
+    function isValid(s: string) {
+        const valid = /^(?:19|20|21)\d{2}-[01]\d$/
+        const m = +s.slice(5)
+        return valid.test(s) && 0 < m && m <= 12;
+    }
```
ludydoo, line 86: Same comment here about dateFns..

ludydoo (changes requested):

web/pwa/package.json
shinroo: Removed it good catch!

pkg/sqlconvert/convert.go
```diff
@@ -174,6 +175,12 @@ func convertFieldToSqlField(formDef *types.FormDefinition, field *types.FieldDef
 		result.DataType = sqlschema.SQLDataType{
 			Date: &sqlschema.SQLDataTypeDate{},
 		}
+	} else if field.FieldType.Month != nil {
+		result.DataType = sqlschema.SQLDataType{
+			VarChar: &sqlschema.SQLDataTypeVarChar{
```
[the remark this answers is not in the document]
shinroo: @nilueps and I implemented this yesterday

web/pwa/src/features/recorder/FieldEditor.tsx
shinroo: I've kept the isValid function in place, but replaced the serialisiation with the format function from `date-fns`

ludydoo (approved):

PR #237, opened 2021-11-03
COR-333: add week field type

shinroo:
- Introduces a new field type `week` which uses the input type `week` and degrades to a regular text input if there is no browser support


web/pwa/src/features/recorder/FieldEditor.tsx
```diff
@@ -71,6 +71,61 @@ export const DateFieldEditor: FC<FieldEditorProps> = props => {
...
+    function getFormattedWeekStringFromDate(date: Date) {
+        const w = getWeekFromDate(date);
+        const y = date.getFullYear();
+        return `${y}-W${w}`
+    }
+
+    function isValid(s: string) {
+        const valid = /^(?:19|20|21)\d{2}-W[0-5]\d$/
+        return valid.test(s) && +s.slice(6) <= 52;
+    }
```
ludydoo, line 103: I think this implementation will not be sufficient.. ISO dates are quite complicated!
Perhaps we could use dateFns ? https://www.npmjs.com/package/date-fns

ludydoo (changes requested):

shinroo:
I've added some unit tests to the `WeekFieldEditor` and `MonthFieldEditor` components, but I am still struggling to get the `jest.fn()` to behave as expected. Other tests are passing.

@ludydoo could you take a look and advise?

ludydoo:
- [] rebase on main
- [] add logic in the `recordStore.Create` to coerce the `Week` values into a valid sql Date. The incoming payload will be like `{"values":{"my_week_field": "2021W23"}}`. The "2021W23" needs to be converted to a valid date (`2021-06-07`)
- [] add logic in the `recordStore.Get/List` to convert the date retrieved from the sql database into `2021W23` format.

internetti:
> * [] rebase on main
> * [] add logic in the `recordStore.Create` to coerce the `Week` values into a valid sql Date. The incoming payload will be like `{"values":{"my_week_field": "2021W23"}}`. The "2021W23" needs to be converted to a valid date (`2021-06-07`)
> * [] add logic in the `recordStore.Get/List` to convert the date retrieved from the sql database into `2021W23` format.

@ludydoo I'm not sure your comment is still true, as far as I can see the date in the database is a proper sql date, please review again

ludydoo:
@internetti the payload from the api is a string
```
values: {"my-week-field":"2021W52"}
```
Though, the database expects a data
```
INSERT into myTable (myWeekField) values ("2021W52") <- I don't think this will work, since it's not a valid date
```
I think we need
here
https://github.com/nrc-no/core/blob/badb3e81092856f363282d82c2074d7742e9d8db/pkg/store/record.go#L323
to convert "2021W52" (payload sent by the js client) to "2021-12-23" (or whatever the first day of the 2021W52 is)
and then, when we retrieve "2021-12-23" from the Database, we need to convert it back to a format that the js client expects ("2021W52")

internetti:
@ludydoo as far as I can see the payload being sent to the api is "2021-12-16T23:00:00.000Z"

internetti:
and the column is a date column, and there is no error when entering a new value by sending a post from the app

ludydoo:
@internetti then we have 2 choices
1. Make the client actually send a proper week like "2021W23" (that's what I thought was the process here..)
2. Ensure that the date stored in the db "2021-10-10T23...." actually is like.. the first day of the week, at 00:00:00 UTC. Because the client might send something like Wednesday at 11:23:40... as a week field value. Which is invalid. It will pose problems if the user defines the week field as a key field.

If a user defines a field as a week fields, and sets this as a key field, the user expects that no 2 records can exist for the same week.
But if the js client sends a "week" value of "Wednesday January the 15th, 2021 at 12:23:22PM" the week is "2021W2". 
Then, the js client sends a second "week" of "Wednesday January the 17th, 2021 at 12:23:22PM". The week is still "2021W2"
But, we should throw here, because there would effectively be 2 records for the same week. Because we didn't "force" the field values to correspond to the first day of the week, like "Jan 03 2021, 00:00:00", we will break the "key" functionality where 2 records cannot have the same week value.

ludydoo:
That's why I would personally prefer that the client sends a proper "2021W21", and that the `record store` converts this to a `sql date` such as `2021-06-05` (or whatever the first day of that week is). The `record store` would also have to do the conversion the other way around, when a user retrieves the record from the db, so that `2021-06-05` is converted back to `2021W21`

ludydoo (approved):

PR #242, opened 2021-11-10
COR-349: seeder uganda

ludydoo:


cmd/seed.go
```diff
@@ -0,0 +1,32 @@
...
+				return err
+			}
+			seeder.Seed(u.Scheme, u.Host)
+			return nil
+		},
+	}
+)
+
+func init() {
+	seedCmd.PersistentFlags().StringVarP(&hostUri, "hostUri", "u", "", "the URI of the Core API, example http://localhost:9000")
```
ludydoo, line 30: rename to host-uri

internetti:
[] rebase
[] remove Subform names, they don't exist anymore
[] use helper methods like buildTextField and buildMultiLineTextField to remove duplication
[] change []*types.FieldDefinition to types.FieldDefinition
[] prevent re-creation of everything if script is run twice

nilueps:
@ludydoo do you know how we could get qualified eyes on this? the excel spreadsheets are a mess so a lot of my work has been a guessing game

PR #244, opened 2021-11-19
COR-355: add validation to databases endpoint

shinroo:
- Introduces a `validation.go` file to the database handler folder, which contains logic to valid `Database` structs
- Introduces unit tests for the validation logic
- Create handler now runs validation after unmarshalling data, returns error if it does not pass

### DIFF
--- a/pkg/server/public/handlers/database/create.go
+++ b/pkg/server/public/handlers/database/create.go
@@ -1,6 +1,9 @@
 package database
 
 import (
+	"fmt"
+	"net/http"
+
 	"github.com/emicklei/go-restful/v3"
 	"github.com/nrc-no/core/pkg/api/meta"
 	"github.com/nrc-no/core/pkg/api/types"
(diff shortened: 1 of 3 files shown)


pkg/server/public/handlers/database/create.go
```diff
@@ -24,6 +26,12 @@ func (h *Handler) Create() http.HandlerFunc {
 			return
 		}
 
+		if !ValidateDBStruct(&db) {
+			l.Error("invalid data")
+			utils.ErrorResponse(w, fmt.Errorf("invalid data"))
+			return
+		}
```
ludydoo, line 33: We already have validation logic in the `pkg/api/types/validation/validation.go`
This should be 

```
if err := validation.ValidateDatabase(&db); err.HasAny(){
  utils.ErrorResponse(w, meta.NewInvalid(..., err)
  return
} 
```

ludydoo (changes requested):

PR #260, opened 2021-12-01
COR-363 add stubs form cases & recipients

ludydoo:


pkg/api/types/recipientdef.go
```diff
@@ -0,0 +1,17 @@
...
+
+	// Extends represents the Parent of the Recipient. Sometimes, we need to have multiple "profiles"
+	// for case recipients. For example, Uganda Beneficiary Individual, Kenya Beneficiary Individual,
+	// Colombia BeneficiaryIndividual are 3 different recipients that "extend" the same "Global Beneficiary Individual"
+	// In this case, Uganda Beneficiary Individual would have Global Beneficiary Individual as Parent
+	//
+	// Also, in some cases, the parent might be a custom FormDefinition that is not a RecipientDef.
+	// If the user has already defined a custom free-form, eg. Individual, and wants to add Case Management
+	// to this currently existing form, then the Extends property would reference the Individual FormDefinition.
+	Extends *FormRef `json:"extends,omitempty"`
```
internetti, line 16: what do you think about naming this "Parent"? I think that way it's clearer what this is

ludydoo: Sounds good :) Just changed it

pkg/api/types/recipientdef.go
```diff
@@ -0,0 +1,17 @@
...
+
+	// Parent represents the Parent of the Recipient. Sometimes, we need to have multiple "profiles"
+	// for case recipients. For example, Uganda Beneficiary Individual, Kenya Beneficiary Individual,
+	// Colombia BeneficiaryIndividual are 3 different recipients that "extend" the same "Global Beneficiary Individual"
+	// In this case, Uganda Beneficiary Individual would have Global Beneficiary Individual as Parent
+	//
+	// Also, in some cases, the parent might be a custom FormDefinition that is not a RecipientDef.
+	// If the user has already defined a custom free-form, eg. Individual, and wants to add Case Management
+	// to this currently existing form, then the Extends property would reference the Individual FormDefinition.
+	Parent *FormRef `json:"extends,omitempty"`
```
neb42, line 16: Why are you renaming this to `extends` instead of `parent`?

PR #266, opened 2021-12-02
Make Recipient form extensible in the future

ludydoo:


pkg/api/types/casedef.go
```diff
@@ -8,5 +18,5 @@ type CaseDef struct {
 	// Recipient represents the type of recipient that this case is for.
 	// For example, if the CaseDef is Colombia Individual Intake, then
 	// the Recipient could be a link to a ColombiaIndividualRecipient form.
-	Recipient FormRef `json:"recipient,omitempty"`
+	Recipient Recipient `json:"recipient,omitempty"`
```
neb42, line 21: Should this not be a `RecipientRef` rather than an actual `Recipient`?

ludydoo: Or `RecipientDef` ?

PR #267, opened 2021-12-02
COR-340 PARTIAL: Setup Linting for Frontend App

nilueps:


internetti (commented):
is there a command to run this manually? and do you think we could set up a prow task to do this before committing?

ludydoo:
> is there a command to run this manually? and do you think we could set up a prow task to do this before committing?

Unfortunately the prow jobs only run after commiting. But the prow job could "add" a new commit itself with linting. Or, it could "fail fast" if the dev forgot to lint.

nilueps:
Added lint scripts, @internetti

PR #268, opened 2021-12-02
COR 340 PARTIAL AddRecordScreen Test

nilueps:
- setup
- setup
- add lint scripts
- setup tests
- split AddRecordScreen into screen and container
- update exports
- add record screen test


web/app/frontend/src/components/form/TextInput.tsx
```diff
@@ -4,7 +4,7 @@ import {Text, TextInput as TextInputRNP} from "react-native-paper";
 import {darkTheme} from "../../constants/theme";
 import {InputProps} from "./FormControl";
 
-const TextInput: React.FC<InputProps> = (
+const TextInput = (
```
internetti, line 7: why this change?

web/app/frontend/src/components/screens/AddRecordScreen.tsx
```diff
@@ -1,106 +1,36 @@
...
+    form,
+    control,
+    onSubmit,
+    formState,
+    isWeb,
+    hasLocalData,
+    isConnected,
+    isLoading,
+}: AddRecordScreenProps)=> {
+  // console.log(form?.fields)
```
internetti, line 30:
```suggestion
```

web/app/frontend/src/components/form/FormControl.tsx
```diff
@@ -8,51 +8,49 @@ import ReferenceInput from "./ReferenceInput";
...
-    value?: any,
-    control: Control<any, object>,
-    errors?: object
+    name: string;
+    fieldDefinition: FieldDefinition;
+    style?: ViewStyle;
+    value?: any;
+    control: Control<any, object>;
+    errors?: object;
+    testID?: string;
```
internetti, line 31: since this is always "formControl", I don't think it needs to be passed as a parameter

web/app/frontend/src/components/screens/AddRecordScreen.tsx
```diff
@@ -1,106 +1,36 @@
...
     return (
         <ScrollView contentContainerStyle={[layout.container, layout.body, common.darkBackground]}>
-
             <View style={[]}>
```
internetti, line 33:
```suggestion
            <View>
```

web/app/frontend/src/components/Router.tsx
```diff
@@ -73,8 +74,8 @@ export default function Router() {
                                 title: routes.addRecord.title
                             }}
                         >
-                            {(props) =>
-                                <AddRecordScreen {...props} state={state} dispatch={dispatch}/>
+                            {({navigation, route}) =>
+                                <AddRecordScreenContainer navigation={navigation} route={route as AddRecordScreenContainerProps['route']} state={state} dispatch={dispatch}/>
```
internetti, line 78: can the linting be set up so each property is on its own line?

web/app/frontend/__tests__/screens/AddRecordScreen.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+    test("renders correctly", () => {
+        const { toJSON } = render(<WrappedComponent />);
+        expect(toJSON()).toMatchSnapshot();
+    });
+
+    test("displays all of a form's fields", () => {
+        const { getAllByTestId } = render(<WrappedComponent />);
+        expect(getAllByTestId(testIds.formControl).length).toEqual(mockForm.fields.length);
+    });
+});
```
internetti, line 57: can you add tests for what is shown when hasLocalData, isConnected, isLoading are true/false?

internetti (changes requested):

web/app/frontend/src/components/form/TextInput.tsx
nilueps: We're not sure, but somehow tsserver was complaining.

Also: https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components

web/app/frontend/src/components/form/FormControl.tsx
nilueps: I believe we may need it because it is being consumed by the Controller component within the FormControl stack

https://react-hook-form.com/api/useform/control

web/app/frontend/src/components/Router.tsx
nilueps: just tested it, it already does that. I just didn't run the fix on the whole package because it would make a massive PR. We should probably submit a dedicated PR for that.
I'll run it for the files in this PR though

web/app/frontend/src/components/form/FormControl.tsx
internetti: yes, it's needed, but it doesn't need to be a parameter because it doesn't change based on the field type. you can just put the constant in line 44 here.

nilueps: I'm not sure I understand. Do you mean `const { control } = useForm()`?

internetti: I mean you can just import the testIDs in this file instead of in the AddRecordScreen and use them in the line `<View style={{ margin: 10 }} testID={testIds.formControl}>`

ludydoo:
@internetti 
Why ? Then all the snapshot tests will fail on the CICD pipeline...

internetti:
> @internetti Why ? Then all the snapshot tests will fail on the CICD pipeline...

didn't think of that, nevermind then :+1:

internetti (dismissed):

web/app/frontend/src/components/form/FormControl.tsx
nilueps: ah, yes you are correct, good catch

web/app/client/src/index.ts
```diff
@@ -1,3 +1,5 @@
-export {default as Client} from "./lib/client"
-export * from "./lib/utils"
-export * from "./lib/types"
+import Client from './lib/client'
+
+export default Client
```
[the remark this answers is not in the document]
nilueps: should be good, although I'm seeing some issues with the `core-auth` imports. might be better to fix this in a separate ticket though

PR #269, opened 2021-12-03
COR-369 additional documentation + minor refactor/renaming + fix go tests

ludydoo:
- Added more documentation for API types
- Renamed `Oauth2Client` to `OAuth2Client`
- Renamed `OAuth2Client.ClientName` to `OAuth2Client.Name`
- Renamed `OAuth2Client.ClientSecret` to `OAuth2Client.Secret`
- Moved the `type formIntf struct{} and type FormInterface interface{}` to a separate file `formintf.go`
- Removed a few unused methods / types
- Fixed the failing backend tests
- Added pipeline to run the go tests automatically


pkg/api/types/formintf.go
```diff
@@ -0,0 +1,56 @@
+package types
```
neb42, line 1: I think it'd be clearer to name this file `forminterface.go` or `form_interface.go`. It's easier to know what it contains rather than having to work out what `intf` refers to.

pkg/api/types/field.go
```diff
@@ -9,26 +9,26 @@ import (
 // A FieldDefinition defines the name, description, boundaries of data collection.
 type FieldDefinition struct {
 	// ID is the ID of the FieldDefinition
-	ID string `json:"id"`
+	ID string `json:"id,omitempty" yaml:"id,omitempty"`
```
neb42, line 12: I don't think we should be setting `omitempty` on all of these fields. Most of them seem to be required fields and should error during read/writes validation. By adding `omitempty` it indicates that a missing value is an expected state and not an error.

pkg/api/types/field_type.go
```diff
@@ -5,21 +5,21 @@ package types
 // cannot have both FieldType.Text and FieldType.Reference defined. Only one is allowed.
 type FieldType struct {
 	// Text represents the configuration for a text field
-	Text *FieldTypeText `json:"text,omitempty"`
+	Text *FieldTypeText `json:"text,omitempty" yaml:"text,omitempty"`
```
neb42, line 8: Same point about `omitempty` in this file

pkg/api/types/field.go
ludydoo: When a user submits a `FormDefinition` with `FieldDefinitions`, in this case the `id` is supposed to be empty, since this generated (server side uuids)...

pkg/api/types/field_type.go
ludydoo: This is the same scenario where only 1 of the fields must be non-nil, all the others must be nil... If we remove `omitempty` the api payload will be like
```
fieldType:
  text: {}
  multilineText: null
  date: null
  ...: null
```
I think we prefer
```
fieldType:
  text: {}
```

pkg/api/types/formintf.go
ludydoo: sounds good

pkg/api/types/field.go
neb42: Ah yes, so in that case I don't think a `FieldDefinition` should have and `ID`. There should be a different struct for when we read from the database.

But I think my point stands for the other fields.

pkg/api/types/field_type.go
neb42: Ah ok. In that case yes it makes sense to omit empty, but I would have structured it slightly differently. Unsure how this would look in go, but in typescript:
```
type FieldType = FieldTypeText | FieldTypeReference | FieldTypeSubForm | FieldTypeMultilineText ....
```
I'm not sure what implications this would have though and isn't a change I'd suggest for this pr.

ludydoo: It's what I usually see on Go for implementing polymorphism, otherwise we would need a discriminator like you suggest, and then we would need custom `MarshalJSON` and `UnmarshalJSON` methods on the `FieldType`. imho this is very standard from working with Kubernetes api a lot, where you get things like
```
- name: my-env-var
  value: VALUE
- name: my-config-var
  valueFrom:
    configMapRef: 
      name: my-config-map
      key: config-key
- name: my-config-var
  valueFrom:
    secretKeyRef: 
      name: my-secret
      key: secret-key
- name: my-field-path-var
  valueFrom:
    fieldRef:
      fieldPath: metadata.namespace
    
```

ludydoo: For now it's a bit weird because we have a lot of empty objects that don't have any config
```
- fieldType:
    text: {}
- fieldType
    multilineText: {}
```
 but we can imagine (in the future)
```
- fieldType:
    text:
      minLength: 10
      maxLength: 20
- fieldType
    multilineText:
      columns: 5
      maxLength: 3000
```

ludydoo: I reviewed the omitempties, removed it for mandatory fields

pkg/api/types/field.go
ludydoo: Just modified for the "required" fields to not have `omitempty`
I think if we have different structs for read & write, this would be another PR entirely, and would be quite big

pkg/api/types/field_type.go
neb42: My issue with this type of polymorphism is that this is valid:
```
- fieldType:
    text:
      minLength: 10
      maxLength: 20
    multilineText:
      columns: 5
      maxLength: 3000
```
And also determining the field type on the client side is more difficult than just doing `fieldType.type`.

ludydoo: look at https://github.com/nrc-no/core/pull/259/files#diff-4788377ddf8ac7c4bd0ae52920cd70b077b6802c71471d4466b670848ac1b900

ludydoo: and https://github.com/nrc-no/core/blob/main/web/app/client/src/lib/utils/getFieldKind.ts

ludydoo: > My issue with this type of polymorphism is that this is valid:

It is valid _json_, but its not a valid `FormDefinition`

neb42: So yeah you can raise errors in the validation, but I prefer having it in the type checking as it's less error prone. But maybe this a better to talk about outside of the pr at some point.

pkg/api/types/field.go
neb42: Yeah, not a suggestion for this pr but I think it's a better pattern generally.

neb42:
lgtm, but I wouldn't mind a second set of eyes particularly around the sql files as I'm not familiar with the codebase yet.

pkg/api/types/oauth2_client.go
```diff
@@ -1,18 +1,36 @@
...
+	// see https://www.oauth.com/oauth2-servers/redirect-uris/
+	RedirectURIs []string `json:"redirectUris"`
+	// AllowedCorsOrigins for this OAuth2Client
+	AllowedCorsOrigins []string `json:"allowedCorsOrigins"`
+	// TokenEndpointAuthMethod fot this OAuth2Client
+	TokenEndpointAuthMethod string `json:"tokenEndpointAuthMethod"`
 }
 
+// Oauth2ClientList represents a list of OAuth2Client
 type Oauth2ClientList struct {
```
nilueps, line 34:
```suggestion
type OAuth2ClientList struct {
```

pkg/sql/convert/convert.go
```diff
@@ -201,11 +174,9 @@ func convertFieldToSqlField(formDef *types.FormDefinition, field *types.FieldDef
 		}
 	} else if field.FieldType.Reference != nil {
 		result.Constraints = append(result.Constraints, sqlschema2.SQLColumnConstraint{
-			Name: fmt.Sprintf("fk__%s_%s_%s__%s_%s_id",
-				formDef.DatabaseID,
+			Name: fmt.Sprintf("fkref__%s_%s__%s_id",
```
nilueps, line 177: is the double "_" intentional?

nilueps (commented):
just the typo, otherwise /lgtm

pkg/sql/convert/convert.go
ludydoo: yes :)

pkg/api/types/oauth2_client.go
ludydoo: done

PR #271, opened 2021-12-03
database store tests and documentation

ludydoo:
Added documentation for the database store
Added tests for the database store, and found a bug or two

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,10 +17,12 @@ require (
 	github.com/gorilla/securecookie v1.1.1
 	github.com/gorilla/sessions v1.2.1
 	github.com/iancoleman/strcase v0.2.0
+	github.com/jackc/pgconn v1.10.1 // indirect
 	github.com/lib/pq v1.10.3
 	github.com/lithammer/shortuuid/v3 v3.0.7
 	github.com/looplab/fsm v0.3.0
 	github.com/manifoldco/promptui v0.9.0
+	github.com/mattn/go-sqlite3 v1.14.9 // indirect
 	github.com/ory/hydra-client-go v1.10.6
 	github.com/rs/cors v1.8.0
 	github.com/satori/go.uuid v1.2.0
@@ -37,5 +39,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 	gorm.io/driver/postgres v1.2.1
 	gorm.io/driver/sqlite v1.2.3
-	gorm.io/gorm v1.22.2
+	gorm.io/gorm v1.22.4
 )
--- a/pkg/store/clear.go
+++ b/pkg/store/clear.go
@@ -2,15 +2,34 @@ package store
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"github.com/nrc-no/core/pkg/logging"
 	"github.com/nrc-no/core/pkg/sql/convert"
 	"go.uber.org/zap"
 	"gorm.io/gorm"
+	"os"
 )
+
+const clearProtectionEnvVar = "VERY_DANGEROUSLY_ALLOW_TO_CLEAR_DB"
+
+func checkClearProtection(ctx context.Context) error {
+	l := logging.NewLogger(ctx)
+	requiredEnvVar, ok := os.LookupEnv(clearProtectionEnvVar)
+	if len(requiredEnvVar) == 0 || !ok || requiredEnvVar != "yes" {
+		l.Error(fmt.Sprintf("Cannot run the Clear method. You must set the %s=yes environment variable\n", clearProtectionEnvVar))
+		return errors.New("required environment variable not present")
+	}
+	return nil
+}
 
 // Clear deletes everything from the database. Useful for development.
 // Be careful in production...
 func Clear(ctx context.Context, db *gorm.DB) error {
+
+	if err := checkClearProtection(ctx); err != nil {
+		return err
+	}
 
 	l := logging.NewLogger(ctx)
 
--- a/pkg/store/clear.go
+++ b/pkg/store/clear.go
@@ -8,6 +8,8 @@ import (
 	"gorm.io/gorm"
 )
 
+// Clear deletes everything from the database. Useful for development.
+// Be careful in production...
 func Clear(ctx context.Context, db *gorm.DB) error {
 
 	l := logging.NewLogger(ctx)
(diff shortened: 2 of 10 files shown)


go.mod
```diff
@@ -17,10 +17,12 @@ require (
 	github.com/gorilla/securecookie v1.1.1
 	github.com/gorilla/sessions v1.2.1
 	github.com/iancoleman/strcase v0.2.0
+	github.com/jackc/pgconn v1.10.1 // indirect
```
internetti, line 20: what does the indirect mean?

ludydoo: It's a dependency that is not directly used by our code, but is required by another dependency (in this case `gorm.io/gorm`, which is our database connection). 
It's needed because this package exposes the errors that are returned by the database, and we need to check the database error and map it to an api error, for example `ErrUniqueConstraint` -> `HTTP 409 Conflict`

ludydoo: It's also just used in our tests

pkg/store/clear.go
```diff
@@ -8,6 +8,8 @@ import (
 	"gorm.io/gorm"
 )
 
+// Clear deletes everything from the database. Useful for development.
+// Be careful in production...
 func Clear(ctx context.Context, db *gorm.DB) error {
```
internetti, line 14: should we add a check for the environment here?

internetti: or create a command line script for it instead of having that in the code

pkg/store/database.go
```diff
@@ -6,36 +6,63 @@ import (
...
 }
 
+type deleteDatabaseFn func(db *gorm.DB, databaseId string) error
+type createDatabaseFn func(db *gorm.DB, database *types.Database) error
+
+// databaseStore is the implementation of DatabaseStore
 type databaseStore struct {
+	// db is the database Factory
 	db Factory
+	// deleteDatabase is a function for deleting the backing sql database
```
internetti, line 54: what does backing mean?

ludydoo: when we create an object `Database`, it creates an actual `SQL Database` on the database server...

pkg/store/clear.go
ludydoo: Yeah, i'll add a `CORE_VERY_DANGEROUSLY_ALLOW_CLEAR_DB` env var, needed to run this

pkg/store/database.go
internetti: could we say "actual" instead of "backing" then? "backing" sounds like "supporting" to me

ludydoo: done :)

internetti:
okay, so this does look good to me, but could someone with more go expertise check this? @nilueps

pkg/store/clear.go
```diff
@@ -2,14 +2,35 @@ package store
...
+	"errors"
+	"fmt"
 	"github.com/nrc-no/core/pkg/logging"
 	"github.com/nrc-no/core/pkg/sql/convert"
 	"go.uber.org/zap"
 	"gorm.io/gorm"
+	"os"
 )
 
+const clearProtectionEnvVar = "VERY_DANGEROUSLY_ALLOW_TO_CLEAR_DB"
```
nilueps, line 14: could we define these in a centralized place?

ludydoo: We could, but this one is really a one-off, and the user will get this message when doing `go run . clear` .. 
Also, it's not clear where we would put them, since we don't use env vars anyway, (we use cobra/viper `yaml` config file)

pkg/store/context.go
```diff
@@ -9,6 +9,21 @@ import (
...
+// storeName is the name of the store, for logging purposes
+// actionName is the name of the action being done, for logging purposes
+// fields is an arbitrary list of fields for logging
+//
+// Returns
+// context.Context the new populated request context with the logger
+// gorm.DB the database connection
+// zap.Logger the logger
+// doneFunc the function that logs the execution time
 func actionContext(ctx context.Context, factory Factory, storeName, actionName string, fields ...zap.Field) (context.Context, *gorm.DB, *zap.Logger, func(), error) {
```
nilueps, line 27:
```suggestion
func actionContext(ctx context.Context, factory Factory, storeName, actionName string, fields ...zap.Field) (context.Context, *gorm.DB, *zap.Logger, doneFunc func(), error) {
```

I think naming the closure will make using the function easier

pkg/store/context.go
```diff
@@ -9,7 +9,22 @@ import (
...
+// storeName is the name of the store, for logging purposes
+// actionName is the name of the action being done, for logging purposes
+// fields is an arbitrary list of fields for logging
+//
+// Returns
+// context.Context the new populated request context with the logger
+// gorm.DB the database connection
+// zap.Logger the logger
+// doneFunc the function that logs the execution time
+func actionContext(ctx context.Context, factory Factory, storeName, actionName string, fields ...zap.Field) (context.Context, *gorm.DB, *zap.Logger, doneFunc func(), error) {
```
nilueps, line 27: turns out we can't mix named and unamed return values
```suggestion
func actionContext(ctx context.Context, factory Factory, storeName, actionName string, fields ...zap.Field) (ctx context.Context, dbConn *gorm.DB, logger *zap.Logger, doneFunc func(), err error) {
```

pkg/store/errors.go
```diff
@@ -0,0 +1,59 @@
...
+type pgErr struct {
+	err *pgconn.PgError
+}
+
+func (s *pgErr) IsErrCode(code string) bool {
+	return s.err.Code == code
+}
+
+func (s *pgErr) IsUniqueConstraintErr() bool {
+	return s.IsErrCode("23505")
```
nilueps, line 41: Where does this number come from?

nilueps (commented):
Looks good in general

pkg/store/errors.go
ludydoo: This is a documented PostgreSQL error code. 
https://www.postgresql.org/docs/13/errcodes-appendix.html

ludydoo: Gonna add this link to the file

pkg/store/clear.go
neb42: I'd be quite nervous leaving this code in when we deploy a production version. A misconfigured environment could wipe the database with no warning.

ludydoo:
@nilueps  if it lgty, you can add the comment, we got some good review on this

pkg/store/errors.go
neb42: May be worth making it a variable

ludydoo: Just did ;)

PR #272, opened 2021-12-03
Tests and documentation for the Folder store

ludydoo:
**Changes to the `store.Suite`**
There was already logic to setup a `store.DatabaseSuite` test suite. I've renamed this to just `store.Suite`. This `store.Suite` is the parent `Suite` for all store tests. 
So there is now `store.Suite` (instead of `store.DatabaseSuite`), and methods like
- `func (s *Suite) TestDatabaseCreate() {}`
- `func (s *Suite) TestFolderCreate() {}` 
- ...

**Changes in this PR**
- Removed unecessary `db.WithContext(ctx)`as this logic is already in the `store.actionContext`, which is the boilerplate code for all store actions.
- Added checks when creating a `Folder` for the existence of the `Database` and `parent` folder (if any), as well as the appropriate error messages
- Fixed error where a foreign key constraint would fail, even if we specify an empty string for the `Folder.parent`. 
- Fixed error where deleting a **parent** folder would fail because the foreign key constraint `OnDelete` was not set to `Cascade`
- Added tests for `store.FolderStore` for `create`, `get`, `list`, `delete`

### DIFF
--- a/pkg/store/database.go
+++ b/pkg/store/database.go
@@ -26,9 +26,9 @@ type DatabaseStore interface {
 // NewDatabaseStore returns a new DatabaseStore
 func NewDatabaseStore(db Factory) DatabaseStore {
 	return &databaseStore{
-		db:             db,
-		deleteDatabase: convert.DeleteDatabaseSchemaIfExist,
-		createDatabase: convert.CreateDatabase,
+		db:                   db,
+		deleteDatabaseSchema: convert.DeleteDatabaseSchemaIfExist,
+		createDatabaseSchema: convert.CreateDatabase,
 	}
 }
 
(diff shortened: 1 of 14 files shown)


pkg/store/database_test.go
```diff
@@ -7,204 +7,174 @@ import (
...
-			return nil
-		},
-		db: d.dbFactory,
-	}
-	d.store = s
-}
-
-func (d *DatabaseSuite) TestCreate() {
+// TestDatabaseCreate tests that we can create a database successfully
+func (d *Suite) TestDatabaseCreate() {
```
nilueps, line 14: some of these receivers are named `d`, while others are named `s`; could we choose one, just for the sake of consistency?

pkg/store/database.go
```diff
@@ -101,7 +101,7 @@ func (d *databaseStore) Delete(ctx context.Context, databaseID string) error {
 	err = db.Transaction(func(tx *gorm.DB) error {
 
 		l.Debug("deleting database schema")
-		if err := d.deleteDatabase(tx, databaseID); err != nil {
+		if err := d.deleteDatabaseSchema(tx, databaseID); err != nil {
 			l.Error("failed to delete database schema", zap.Error(err))
 			return meta.NewInternalServerError(err)
```
nilueps, line 106: Not a big deal, by why are these status and error messages store in a package called `meta`. It's a little confusing IMHO.

ludydoo: It's like.. meta stuff for the api, like status code, error codes, that kind of stuff

pkg/store/database_test.go
ludydoo: Nice find, will update

nilueps (approved):
All tests passing on my end

PR #273, opened 2021-12-04
Introduce SQL Manager

ludydoo:
## SQL Manager

The main complexity of the core backend is the conversion of a `FormDefinition` into actual `SQL Tables`. Especially since a `FormDefinition` can have `SubForms`, or `References`. We must create the appropriate tables, fields, foreign keys, constraints that match the user-defined `FormDefinition`.

The current implementation of converting a `FormDefinition` into multiple `SQLTable, SQLField, SQLTableConstraint` is badly implemented. It was done quick and dirty in order to get something out as quick as possible.

**This PR does not switch the current `store.FormStore` implementation. It rather inserts this new package in the codebase for review, as well as introduce other changes to the codebase that offer a path forward**

This PR introduces a new package `sqlmanager` that simplifies part of the process, both by reducing the complexity of the implementation, as well as combining the logic that was beforehand scattered around the codebase.

### How it works
1. The `SqlManager.PutForms(myFormDefinition)` is called
2. The `SqlManager` inspects all the aspects of the given `FormDefinition`. The fields, name, etc. It builds a list of `SQL Statements` (`create table myTable`, `alter table add myColumn`, ...), that are needed to create the actual SQL table from the given form definition
3. Currently, only creating new forms is supported, but it is designed to support updates as well in the future.

#### Remove `FieldTypeSubForm` `ID`, `Code` and `Name`
```
Currently
FormDefinition
  Fields:
    - ID: myField
      Name: My Field
      Code: MY_FIELD
      FieldType:
        SubForm:
          Name: My Sub Form
          ID: mySubFormID
          Code: MySubFormCode
          Fields: [...]
---
This PR
FormDefinition
  Fields:
    - ID: myField
      Name: My Field
      Code: MY_FIELD
      FieldType:
        SubForm:
          Fields: [...]
```
Before, the name of a sub form field could be different from the name of the subform.  This simplifies a lot the `FormInterface`

This will also simplify the frontend quite a lot, as currently we need to do a quite complex lookup on the frontend to correlate a sub form record with the parent record. 

#### Rename `Record.ParentID` to `Record.OwnerID`. 

It is more representative that this represents a sub form record. The `ParentID` would be conflicting with the notion of inheritance that we want to introduce in the cases.

#### Add some helper methods
- Created `SQLTables` which is a list of `SQLTable`
- Created `SQLFields` which is a list of `SQLField`
- Created `SQLTableConstraints` which is a list of `SQLTableConstraint`

#### Add typed errors on the `sqlschema` package
Errors such as 
- `ColumnNotFoundErr`
- `TableNotFoundErr` 

#### Fix frontend imports of the core-api-js-client

There was a duplication of code for the core-js-api-client.
The types were defined in both the `core-api-js-client` and the `pwa` app
This removes the `pwa` api types and use the `core-api-js-client` types instead
It also fixes the imports `import {FormDefinition} from "core-api-js-client/lib/types"` -> `import {FormDefinition} from "core-api-js-client"`


pkg/sql/schema/sql_table_constraint.go
```diff
@@ -64,7 +75,7 @@ func NewForeignKeyTableConstraint(
 
 func (c SQLTableConstraint) DDL() DDL {
 	ddl := DDL{}
-	ddl.WriteF("constraint %s", pq.QuoteIdentifier(c.Name))
+	ddl = ddl.WriteF("constraint %s", pq.QuoteIdentifier(c.Name))
```
ludydoo, line 78: this was a bug

pkg/sql/schema/sql_table_test.go
```diff
@@ -19,7 +19,7 @@ func TestSQLTable_DDL(t *testing.T) {
 		{
 			name:   "empty",
 			fields: args{Name: "empty"},
-			want:   NewDDL(`create table "public"."empty";`),
+			want:   NewDDL(`create table "public"."empty"();`),
```
ludydoo, line 22: this is a valid syntax to create an empty sql table (with no columns)

.prow/push-core-go-test.yaml
```diff
@@ -8,7 +8,19 @@ presubmits:
           command: [ "bash", "-c" ]
           args:
             - >
-              go mod download &&
-              make gen &&
-              git diff --exit-code &&
-              go test ./pkg/...
+              echo "Creating user" &&
```
ludydoo, line 11: i installed `embedded-postgresql` go package for tests. Though, it's impossible to run as root, hence the modification here.

pkg/api/types/field.go
```diff
@@ -32,86 +27,23 @@ type FieldDefinition struct {
...
 	return f.IsKind(FieldKindReference)
 }
 
 // IsSubFormField returns whether a the field is a FieldTypeSubForm or not
-func (f FieldDefinition) IsSubFormField() bool {
+func (f FieldDefinition) IsSubFormField() (bool, error) {
 	return f.IsKind(FieldKindSubForm)
 }
 
-// GetKind returns the FieldKind of the field
```
ludydoo, line 49: All of this logic already exists on `FieldType`

pkg/api/types/field.go
```diff
@@ -32,86 +27,23 @@ type FieldDefinition struct {
...
-}
-
-// FieldIDs returns the of FieldDefinition.ID for all the FieldDefinitions
-func (f FieldDefinitions) FieldIDs() sets.String {
-	result := sets.NewString()
-	for _, definition := range f {
-		result.Insert(definition.ID)
-	}
-	return result
-}
```
ludydoo, line 117: All of this was unused

pkg/api/types/field_type.go
```diff
@@ -83,19 +105,8 @@ type FieldTypeSingleSelect struct {
 // For example, given a form "Projects", this form could have a subform "Monthly Deliveries".
 // The "Monthly Deliveries". There could be multiple "Monthly Deliveries" for a single "Project".
 type FieldTypeSubForm struct {
-	// ID represents the ID of the sub form
-	ID string `json:"id" yaml:"id"`
-	// Name represents the Name of the sub form
-	Name string `json:"name" yaml:"name"`
-	// Code represents the unique Code for the subform Field
-	Code string `json:"code,omitempty" yaml:"code,omitempty"`
```
ludydoo, line 91: We remove subform id, name and code. Only `fields` is now part of the `SubForm`. It will inherit the name and if from the `field`

pkg/constants/constants.go
```diff
@@ -7,16 +7,4 @@ const (
...
-	QueryParamCode  = "code"
-
-	SessionState        = "state"
-	SessionIDToken      = "idToken"
-	SessionRefreshToken = "refreshToken"
-	SessionAccessToken  = "accessToken"
-	SessionTokenExpiry  = "tokenExpiry"
-	SessionTokenType    = "tokenType"
-	SessionProfile      = "profile"
-	SessionDesiredURL   = "desiredUrl"
```
ludydoo, line 21: All unused stuff

pkg/sql/schema/sql_field.go
```diff
@@ -14,74 +14,15 @@ type SQLField struct {
 	Comment     string                `json:"comment,omitempty" yaml:"comment,omitempty"`
 }
 
-func NewSQLField(name string) SQLField {
```
ludydoo, line 17: All unused..

pkg/sql/schema/sql_table.go
```diff
@@ -74,7 +93,7 @@ func (s SQLTable) DDL() DDL {
 		statements = append(statements, field.DDL().WriteBefore("  "))
 	}
 	for _, constraint := range s.Constraints {
-		statements = append(statements, constraint.DDL().WriteBeforeF("  constraint %s", pq.QuoteIdentifier(constraint.Name)))
+		statements = append(statements, NewDDL("  ").Merge(constraint.DDL()))
```
ludydoo, line 96: this was also a workaround for the bug that was fixed in the `SQLTableConstraint.DDL`
No need for the workaround since the bug is fixed

pkg/api/types/validation/form.go
```diff
@@ -54,7 +57,7 @@ var (
 
 func init() {
 	validateFormNameFn = ValidateFormName
-	validateDatabaseIdFn = ValidateDatabaseName
+	validateDatabaseIdFn = ValidateFormDatabaseID
```
ludydoo, line 60: This was a bug

pkg/api/types/validation/form.go
```diff
@@ -234,8 +264,6 @@ func ValidateFieldTypeText(ftText *types.FieldTypeText, path *validation.Path) v
 
 func ValidateFieldTypeSubForm(ftSF *types.FieldTypeSubForm, path *validation.Path) validation.ErrorList {
 	var result []*validation.Error
-	result = append(result, ValidateFormName(ftSF.Name, path.Child("name"))...)
```
ludydoo, line 237: sub forms dont have name anymore

PR #275, opened 2021-12-05
add some tests for utility methods

ludydoo:


cmd/devinit/utils.go
```diff
@@ -218,10 +218,16 @@ func getOrCreateRandomSecret(length int, filePaths ...string) ([]byte, error) {
...
 
 	if exists {
-		return os.ReadFile(filePath)
+		fileContent, err := os.ReadFile(filePath)
+		if err != nil {
+			return nil, err
+		}
+		if len(fileContent) != 0 {
+			return fileContent, nil
+		}
```
neb42, line 227: Should this return something if `len(fileContent) == 0`? Previously it would return here, but now it will continue.

ludydoo:
That would fix a bug where if some reason the file was created but is empty, this would recreate the file. We had this bug before, this would fix it 

Envoyé de mon iPhone

PR #276, opened 2021-12-06
COR-356 Add validation for folders

ludydoo:


pkg/server/public/handlers/folder/create.go
```diff
@@ -25,6 +27,13 @@ func (h *Handler) Create() http.HandlerFunc {
 
 		folder.ID = uuid.NewV4().String()
 
+		if errList := validation.ValidateFolder(&folder); errList.HasAny() {
```
ludydoo, line 30: Change HasAny to !IsEmpty

internetti (commented):
Is there a way to adapt the error messages? when the folder name is too short, the message is "Too long: must have at minimum 3 bytes", I think "characters" would be better than "bytes"

ludydoo:
@internetti we would have to modify the `validation.TooLong` / `validation.TooShort` methods. Perhaps we could tackle this in another tix ?

PR #277, opened 2021-12-06
COR-357 Add validation for records

ludydoo:


pkg/server/public/handlers/record/create.go
```diff
@@ -29,8 +31,17 @@ func (h *Handler) Create() http.HandlerFunc {
 			utils.ErrorResponse(w, err)
 			return
 		}
+
 		input.DatabaseID = form.DatabaseID
 
+		l.Debug("validating record")
+		if validationErrs := validation.ValidateRecord(&input, form); validationErrs.HasAny() {
```
ludydoo, line 38: Change to !IsEmpty. Check for other instances of HasAny. Fix naming of HasAny since it's misleading. Perhaps ´HasSome ´

pkg/server/public/handlers/record/update.go
```diff
@@ -41,6 +42,22 @@ func (h *Handler) Update(recordId string) http.HandlerFunc {
...
+		l.Debug("getting form for record", zap.String("form_id", input.FormID))
+		form, err := h.formStore.Get(req.Context(), input.FormID)
+		if err != nil {
+			l.Error("failed to get record form", zap.Error(err))
+			utils.ErrorResponse(w, err)
+			return
+		}
+
+		l.Debug("validating record")
+		if validationErrs := validation.ValidateRecord(&input, form); validationErrs.HasAny() {
```
ludydoo, line 54: Same comment for HasAny

PR #283, opened 2021-12-08
Add API Tests to Core

ludydoo:
This PR sets up some basic API tests to be run
It ensures that we can do basic operations such as
- Creating, Getting, Listing databases
- Creating, Getting, Listing folders
- Creating, Getting, Listing forms
- Creating, Getting, Listing records

We use `embedded-postgres` to run an actual in-process database. Since `go test` does not run all tests on the same process, we must synchronize setup/teardown of the database using a file lock.


pkg/server/public/tests/form_test.go
```diff
@@ -0,0 +1,189 @@
...
+		}, {
+			name: "With Required Reference Field",
+			fields: tu.Fields(
+				tu.AReferenceField(otherForm, tu.FieldName("My Field"), tu.FieldRequired(true)),
+			),
+		}, {
+			name: "With Key Reference Field",
+			fields: tu.Fields(
+				tu.AReferenceField(otherForm, tu.FieldName("My Field"), tu.FieldRequired(true), tu.FieldKey(true)),
+			),
```
nilueps, line 139: Add cases for broken inputs. i.e.: missing required fields, wrong types, etc

pkg/sqlmanager/schema_test.go
```diff
@@ -1,50 +1,65 @@
...
+	exitVal := m.Run()
+	done()
+	os.Exit(exitVal)
+}
+
 func (s *Suite) SetupSuite() {
 
-	pg := embeddedpostgres.NewDatabase()
-	if err := pg.Start(); !assert.NoError(s.T(), err) {
+	sqlDb, err := gorm.Open(postgres.Open("host=localhost port=15432 user=postgres password=postgres dbname=postgres sslmode=disable"))
```
nilueps, line 45: can we set the port dynamically?

nilueps (commented):
Clean & Snappy

pkg/server/public/tests/form_test.go
ludydoo: Yeah... this part was just to test the rest api, make sure that we can read and write without breaking anything. Def. would need more work in the future, more cases, etc.

pkg/sqlmanager/schema_test.go
ludydoo: Yeah we could add an env var like TEST_POSTGRES_PORT=... and read from it, with a default

ludydoo:
@nilueps  lgtm ?

pkg/sqlmanager/schema_test.go
nilueps: Yea

nilueps:
there's a conflict

PR #284, opened 2021-12-08
add prow job for pwa tests and fix those tests

internetti:


.prow/pull-core-pwa-test.yaml
```diff
@@ -0,0 +1,15 @@
+postsubmits:
+  - name: pull-core-pwa-test
+    run_if_changed: "^web/pwa"
```
ludydoo, line 3: This probably needs to run if web/pwa, web/auth or web/app/client changed. I think there is another prow job with that kind of regex that you could use as inspiration

ludydoo: Lol, now that we merged the lerna stuff, this probably still need to change ^^

ludydoo: After all, I think that the pwa job can be removed, I believe that the changes to lerna basically test all frontend stuff in a single job

ludydoo (approved):

PR #285, opened 2021-12-08
COR 340 PARTIAL ViewRecordScreenContainer test

nilueps:
- separate screen and container
- rename
- container
- add tests


web/app/frontend/__tests__/screen_containers/ViewRecordScreenContainer.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+    formsById: {},
+};
+
+mockState.formsById[mockFormId] = {
+    recordsById: {},
+};
+
+const mockValues = Array(10).fill("a value");
+
+mockState.formsById[mockFormId].recordsById[mockRecordId] = { values: mockValues };
```
ludydoo, line 26: values are now an array like
```
values:
- fieldId: abc
  value: value123
```

nilueps: LIke this?

PR #287, opened 2021-12-08
use new SQLManager and simplify form store implementation

ludydoo:
Switched the `store.FormStore` implementation to use the new `sqlmanager`. 
Also, removed the weird `FormHierarchy` thingy, and use a much simpler and extensible implementation.


pkg/api/types/form.go
```diff
@@ -36,13 +36,8 @@ func (f *FormDefinition) GetFields() FieldDefinitions {
...
-func (f *FormDefinition) GetOwner() FormInterface {
-	return nil
-}
-
-// HasOwner implements FormInterface.HasOwner
-func (f *FormDefinition) HasOwner() bool {
+// IsSubForm implements FormInterface.IsSubForm
+func (f *FormDefinition) IsSubForm() bool {
 	return false
 }
```
nilueps, line 42: Why does this always return false?

pkg/sql/convert/create.go
```diff
@@ -1,11 +1,10 @@
 package convert
 
 import (
-	"context"
 	"fmt"
 	"github.com/lib/pq"
 	"github.com/nrc-no/core/pkg/api/types"
-	"github.com/nrc-no/core/pkg/sql/schema"
+	sqlschema2 "github.com/nrc-no/core/pkg/sql/schema"
```
nilueps, line 7:
```suggestion
	"github.com/nrc-no/core/pkg/sql/schema"
```
... and rename the rest

nilueps (commented):
Are we entirely eliminating the concept of ownership for FormDefinitions?

nilueps (approved):
/lgtm

PR #295, opened 2021-12-09
Eslint fix

neb42:
* Couple of linting config updates
* Run eslint --fix to get the easy linting wins
* Plenty of errors remain, but can be fixed graduallyb


internetti:
rebase needed

internetti:
sonar is complaining about the http example url used in some tests, needs to be replaced by https ...

internetti:
you'll need to do another rebase, it did complain about the App.tsx test too

PR #301, opened 2021-12-10
Lerna tweaks

neb42:
* Finishing tweaks to lerna configuration
* Cleaned up some build scripts/tsconfig
* Moved local-odic-provider to yarn from npm
* Fixed a bad usage of a hook in native app


nilueps (commented):
/lgtm

PR #303, opened 2021-12-13
COR-406 Core server Dockerfile

ludydoo:


build/package/core.Dockerfile
```diff
@@ -1,29 +1,25 @@
...
+ADD . .
 
-# Build the application
-RUN go build -o main .
+RUN go mod verify
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o main .
 
 # Build a small image
-FROM scratch
+FROM golang:alpine
```
neb42, line 20: Why do we need golang here? Can we not run the binary with go?

ludydoo: Yeah, good catch. Had some troubles with compiling in sqlite binaries that we use in tests

PR #304, opened 2021-12-13
COR-412 Core admin Dockerfile

ludydoo:


configs/authnz-frontend/nginx.conf.template
```diff
@@ -0,0 +1,45 @@
...
+
+  index index.html;
+
+  location / {
+       try_files $uri $uri/ /index.html;
+  }
+
+  location ~ $/(static)/ {
+    gzip on;
+    gzip_http_version  1.0;
```
neb42, line 16: I'm not sure what difference this would have, but why isn't this `1.1`?

ludydoo: Will fix. Seems 1.1 is the default, so I'll remove this stanza

neb42:
Looks good for now, but we may want to move the package building out of here and into a build/publish to npm pipeline. This can then just install/build the react app.

PR #305, opened 2021-12-13
COR-411 Core pwa Dockerfile

ludydoo:


neb42:
Looks good for now, but we may want to move the package building out of here and into a build/publish to npm pipeline. This can then just install/build the react app.

PR #309, opened 2021-12-13
COR-415 remove urls.self server option

ludydoo:
The urls.self was ignored (overwritten) anyways

### DIFF
--- a/cmd/devinit/core.go
+++ b/cmd/devinit/core.go
@@ -24,16 +24,9 @@ func (c *Config) makeCore() error {
 		database: c.coreDbName,
 	})
 
-	const proxyUrl = "https://localhost:8443"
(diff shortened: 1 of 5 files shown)


cmd/devinit/core.go
```diff
@@ -24,16 +24,9 @@ func (c *Config) makeCore() error {
 		database: c.coreDbName,
 	})
 
-	const proxyUrl = "https://localhost:8443"
-
 	coreConfig := map[string]interface{}{
 		"serve": map[string]interface{}{
 			"admin": map[string]interface{}{
-				"cache": map[string]interface{}{
```
ludydoo, line 32: Also, the admin server does not need a redis cache config

PR #310, opened 2021-12-13
COR-409 fix auth token injection logic

ludydoo:
We need to be able to inject either the `id_token` or the `access_token`. 
This PR introduces an option to inject either

### DIFF
--- a/frontend/apps/admin/src/client/client.ts
+++ b/frontend/apps/admin/src/client/client.ts
@@ -1,4 +1,4 @@
-import axios, { AxiosError, AxiosInstance, Method } from 'axios';
+import { AxiosError, AxiosInstance, Method } from 'axios';
 
 import { Client as ApiClient } from '../types/types';
 import { clientResponse } from '../utils/responses';
@@ -28,12 +28,16 @@ import {
 import { RequestOptions, Response } from '../types/utils';
 
 export default class Client implements ApiClient {
+  public axiosInstance: AxiosInstance;
+
   private adminv1 = 'apis/admin.nrc.no/v1';
 
-  public constructor(
-    public readonly address = 'https://localhost:9001/',
-    public readonly axiosInstance: AxiosInstance = axios.create(),
-  ) {}
+  private readonly address: string;
+
+  public constructor(address: string, axiosInstance: AxiosInstance) {
+    this.address = address;
+    this.axiosInstance = axiosInstance;
+  }
 
   do<TRequest, TBody>(
     request: TRequest,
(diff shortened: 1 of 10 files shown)


frontend/apps/admin/src/client/client.ts
```diff
@@ -30,10 +30,14 @@ import { RequestOptions, Response } from '../types/utils';
...
 
-  public constructor(
-    public readonly address = 'https://localhost:9001/',
-    public readonly axiosInstance: AxiosInstance = axios.create(),
-  ) {}
+  private readonly address: string;
+
+  private axiosInstance: AxiosInstance;
+
+  public constructor(address: string, axiosInstance: AxiosInstance) {
```
ludydoo, line 37: eslint complained about a useless constructor

frontend/apps/admin/src/client/client.ts
```diff
@@ -50,7 +54,7 @@ export default class Client implements ApiClient {
       headers = options?.headers;
     }
 
-    return axios
+    return this.axiosInstance
```
ludydoo, line 57: Here, we were ignoring the passed `axiosInstance` entirely. So injecting the token as an interceptor had no effect

frontend/apps/admin/src/hooks/hooks.tsx
```diff
@@ -43,7 +45,10 @@ export function useIdentityProviders(organizationId: string): IdentityProvider[]
   return idps;
 }
 
-export function useFormValidation<T extends object = { [key: string]: any }>(isNew: boolean, form: UseFormReturn<T>) {
+export function useFormValidation<T extends Record<string, unknown> = { [key: string]: any }>(
```
ludydoo, line 48: eslint complained that we should use `Record`

frontend/packages/core-auth/src/components/AuthWrapper.tsx
```diff
@@ -17,8 +17,10 @@ const AuthWrapper: React.FC<AuthWrapperProps> = ({
   redirectUri,
   customLoginComponent,
   handleLoginErr = console.log,
+  injectToken = 'access_token',
 }) => {
-  const browser = new Browser();
+  const browser = useMemo(() => new Browser(), []);
```
ludydoo, line 22: Here, a new `Browser` was re-created over and over

frontend/packages/core-auth/src/components/AuthWrapper.tsx
```diff
@@ -130,7 +149,7 @@ const AuthWrapper: React.FC<AuthWrapperProps> = ({
         {customLoginComponent ? (
           customLoginComponent({ login: handleLogin })
         ) : (
-          <button onClick={handleLogin} role="button">
+          <button type="button" onClick={handleLogin}>
```
ludydoo, line 152: eslint complained that we should specify the button type

PR #320, opened 2021-12-14
Remove logrus logger

ludydoo:
As I was deploying core onto Backbone, I came over some strange `too many open files` issue. I tenatively removed logrus (which was supposed to be removed anyways) to make sure that didn't cause the issue. It did not, but I'm still issuing this PR to remove it.
Also, this adds a retry around verifying the redis connection on startup.


internetti:
feel like the redis thing could have been it's own PR, since it's only 10 files, I'll let it slide :P
/lgtm

PR #321, opened 2021-12-14
parallel authentication and authorization on authnz-bouncer

ludydoo:


internetti (commented):
isn't authorization something that only makes sense for authenticated users? does it make sense to parallelize this?

ludydoo:
> isn't authorization something that only makes sense for authenticated users? does it make sense to parallelize this?

Yeah, both the authorizer and the authenticator will "fail fast" if no bearer token is present in the request body

neb42:
I think two failed requests isn't too much extra vs the benefits of speeding up our happy path. In the event of a ddos I don't think a 2x increase in requests is that bad? But is it worth asking github about?

ludydoo:
I believe DDOS could be somewhat mitigated by adding rate limiting on the gateway level. For now, `authorize` and `authenticate` don't do much. They just call hydra. But `authorize` will do more stuff in the future, like calling the (yet-to-be) zanzibar shaenigans and/or `OPA`. Also, if a token is not present in the request, there will be no calls at all, as there's a check in the authorizer/authenticator for the presence of a `Bearer` token

ludydoo:
@lojikil mentions that an attacker could get a different response given either authn/authnz fail, so could get more information on the system/token.
In this case, we would probably need to perform authn and authz (can be in parallel), but we do not write any response until both gates have passed, and the response is deterministic and not reliant on which of the two goroutines finish first.

PR #323, opened 2021-12-14
Cor 427 reuse sqlmanager logic for writing records

ludydoo:
We had quite a lot of work done to the `sqlManager`. This is now very much compatible with writing records as well.
The `sqlmanager.Interface` had a method `PutForms(forms, ...)`
This introduces `PutRecords(...)`
So that we can re-use the sql reconciliation logic/sql statement generator logic for both form creation and record creation.

It makes sense to do it as part of the MultiSelect field, because the previous field types did not require inserting rows into multiple columns at once. Since the `MultiSelect` field type stores data in an additional association table, it would have created a lot of duplicated code with the `sqlManager`.


pkg/sqlmanager/writer.go
```diff
@@ -0,0 +1,281 @@
+package sqlmanager
```
ludydoo, line 1: This file was renamed from `schema.go` to `writer.go`

PR #325, opened 2021-12-14
Cor 335 multiple selection field

ludydoo:
## COR-426

To allow multi-select field values through the api, we need to allow the user to `POST` either strings or string arrays
so that
```
POST /records

formId: form1
values:
- fieldId: field1
  value: value1
```
Could also allow
```
POST /records

formId: form1
values:
- fieldId: field1
  value: value1
- fieldId: field2
  value: [value2, value3]
```

Current approach introduces a struct `StringOrArray` that will be properly de-serialized based on the json input.

## COR-427

We had quite a lot of work done to the `sqlManager`. This is now very much compatible with writing records as well.
The `sqlmanager.Interface` had a method `PutForms(forms, ...)`
This introduces `PutRecords(...)`
So that we can re-use the sql reconciliation logic/sql statement generator logic for both form creation and record creation.

It makes sense to do it as part of the MultiSelect field, because the previous field types did not require inserting rows into multiple columns at once. Since the `MultiSelect` field type stores data in an additional association table, it would have created a lot of duplicated code with the `sqlManager`. 

## COR-420

Adds logic to generate the sql statements necessary to store the multi select field data


pkg/api/types/record_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"encoding/json"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+// TestStringOrArrayMarshaling tests that we can successfully Marshal and Unmarshal
+// a StringOrArray to/from a string or an array of string, depending on the underlying value.
+func TestStringOrArrayMarshaling(t *testing.T) {
```
nilueps, line 11: Add tests for failing cases?

pkg/sqlmanager/reader.go
```diff
@@ -139,13 +212,13 @@ func readInRecordField(
 	case types.FieldKindMonth, types.FieldKindDate, types.FieldKindWeek:
 		readInDateField(record, field, value, fieldKind)
```
nilueps, line 213: missing return statement?

nilueps (commented):
looks alright to me, but my brain is mush this morning so I'm reluctant to approve, needs another pair of eyes

pkg/sqlmanager/reader.go
ludydoo: done

frontend/apps/pwa/src/components/RecordPicker.tsx
```diff
@@ -7,8 +7,8 @@ import { selectFormOrSubFormById, selectRootForm } from '../reducers/form';
 
 export type RecordPickerProps = {
   disabled?: boolean;
-  recordId: string | undefined;
-  setRecordId: (recordId: string | undefined) => void;
+  recordId: string | null;
```
nilueps, line 10:
```suggestion
  recordId?: string;
```

frontend/apps/pwa/src/components/RecordPicker.tsx
```diff
@@ -7,8 +7,8 @@ import { selectFormOrSubFormById, selectRootForm } from '../reducers/form';
...
+  setRecordId: (recordId: string | null) => void;
```
nilueps, line 11:
```suggestion
  setRecordId: (recordId?: string) => void;
```

frontend/apps/pwa/src/components/RecordPicker.tsx
```diff
@@ -29,16 +29,20 @@ export const RecordPicker: FC<RecordPickerProps> = (props) => {
...
       </select>
     </div>
   );
 };
 
 export type RecordPickerContainerProps = {
-  recordId: string | undefined;
-  setRecordId?: (recordId: string | undefined) => void;
+  recordId: string | null;
+  setRecordId?: (recordId: string | null) => void;
```
nilueps, line 45:
```suggestion
  recordId?: string;
  setRecordId?: (recordId?: string) => void;
```

frontend/apps/pwa/src/components/RecordPicker.tsx
```diff
@@ -50,7 +54,7 @@ export const RecordPickerContainer: FC<RecordPickerContainerProps> = (props) =>
   const dispatch = useAppDispatch();
 
   const form = useAppSelector((state) => {
-    return selectFormOrSubFormById(state, props.formId ? props.formId : '');
+    return selectFormOrSubFormById(state, formId || '');
```
nilueps, line 57:
```suggestion
    return selectFormOrSubFormById(state, formId ?? '');
```

frontend/apps/pwa/src/components/RecordPicker.tsx
```diff
@@ -80,12 +84,12 @@ export const RecordPickerContainer: FC<RecordPickerContainerProps> = (props) =>
   });
 
   const callback = useCallback(
-    (recordId: string | undefined) => {
+    (recId: string | null) => {
```
nilueps, line 87:
```suggestion
    (recId?: string) => {
```

frontend/apps/pwa/src/components/RecordPicker.tsx
ludydoo: the `?` accepts undefined right? I'd like to coerce to null, since that's what we gonna send from the api... Whatdya think?

frontend/apps/pwa/src/components/RecordPicker.tsx
ludydoo: this is eslint automatic stuff...

frontend/apps/pwa/src/components/RecordPicker.tsx
```diff
@@ -98,16 +102,11 @@ export const RecordPickerContainer: FC<RecordPickerContainerProps> = (props) =>
...
-          if (field.key) {
-            result += r.values.find((v: any) => v.fieldId === field.id);
-          }
-        }
-        return result;
+        const result = form?.fields.reduce((prev, next) => {
+          const fieldVal = r.values.find((v) => v.fieldId === next.id)?.value;
+          return fieldVal ? `${prev} ${fieldVal}` : prev;
+        }, '');
+        return result || '';
```
nilueps, line 109:
```suggestion
        return result ?? '';
```

frontend/apps/pwa/src/features/recorder/FieldEditor.tsx
```diff
@@ -1,26 +1,57 @@
...
 import { RecordPickerContainer } from '../../components/RecordPicker';
 
 import { FormValue } from './recorder.slice';
 
 export type FieldEditorProps = {
   field: FieldDefinition;
-  value: any;
-  setValue: (value: any) => void;
+  value: string | string[] | null;
+  setValue: (value: string | string[] | null) => void;
```
nilueps, line 11:
```suggestion
  value?: string | string[];
  setValue?: (value: string | string[]) => void;
```

frontend/apps/pwa/src/features/recorder/FieldEditor.tsx
```diff
@@ -1,26 +1,57 @@
...
+};
+
+export const mapSelectOptions = (required: boolean, key: boolean, options?: SelectOption[]) => {
+  if (!options) {
+    return <></>;
+  }
+  return (
+    <>
+      <option aria-label="no value" disabled={required || key} value="" />
+      {options.map((o) => (
```
nilueps, line 38:
```suggestion
      {options?.map((o) => (
```

frontend/apps/pwa/src/features/recorder/FieldEditor.tsx
```diff
@@ -174,19 +196,33 @@ export const SingleSelectFieldEditor: FC<FieldEditorProps> = (props) => {
...
   return (
     <div className="form-group mb-2">
-      <label className="form-label opacity-75" htmlFor={field.id}>
-        {field.name}
-      </label>
+      {mapFieldLabel(field)}
+      <select
+        className="form-control bg-dark text-light border-secondary"
+        id={field.id}
+        value={value || ''}
```
nilueps, line 203:
```suggestion
        value={value ?? ''}
```

frontend/apps/pwa/src/features/recorder/FieldEditor.tsx
```diff
@@ -174,19 +196,33 @@ export const SingleSelectFieldEditor: FC<FieldEditorProps> = (props) => {
...
+
+export const MultiSelectFieldEditor: FC<FieldEditorProps> = (props) => {
+  const { field, value, setValue } = props;
+  return (
+    <div className="form-group mb-2">
+      {mapFieldLabel(field)}
       <select
         className="form-control bg-dark text-light border-secondary"
         id={field.id}
         value={value || ''}
```
nilueps, line 221:
```suggestion
        value={value ?? ''}
```

nilueps (changes requested):

frontend/apps/pwa/src/components/RecordPicker.tsx
nilueps: I see, in that case better leave it as is, but it means that we will have to explicitely declare the prop as `null` anytime there is no value, I case that's the behavior you want?

ludydoo: I don't know, what do you think @internetti @neb42

internetti: yeah, it is less convenient to explicitly declare null everytime, but if that's what comes from the api, that's what we'll have to do.
we could write mappers that could transform any api payload into a form more convenient for the frontend, if the need for that becomes high enough, but I don't see that yet.

ludydoo: Isn't there something with the `<input>` that we need to pass it a `null` or a `string`, otherwise it complains about being controller/uncontrolled ?

neb42: I'd say `string | null` because `recordId` should default to something and that should be either `null` or `string`. I don't see where it would be `undefined`

frontend/apps/pwa/src/features/recorder/FieldEditor.tsx
ludydoo: This will not happen as we have a check above for falsy options

frontend/apps/pwa/src/features/browser/FormBrowser.tsx
```diff
@@ -113,6 +113,16 @@ function mapRecordCell(field: FieldDefinition, record: Record, getSubFormCount:
     return <td key={field.id}>{value}</td>;
   }
 
+  if (field.fieldType.multiSelect) {
+    const selected = field.fieldType.multiSelect.options.filter((o: SelectOption) => {
+      if (fieldValue == null || fieldValue.value == null) {
+        return false;
+      }
+      return fieldValue.value.includes(o.id);
```
ludydoo, line 121: could this be replaced with `return fieldValue?.value?.includes(o.id)` ?

pkg/server/formsapi/tests/record_test.go
```diff
@@ -40,48 +38,36 @@ func (s *Suite) TestRecordCreateReadList() {
...
+			}, tu.FieldName(singleSelectFieldName)),
+			tu.AMultiSelectField([]*types.SelectOption{
 				{Name: "option1"},
 				{Name: "option2"},
-			},
-				tu.FieldName(singleSelectRequiredFieldName),
-				tu.FieldRequired(true),
-			),
+				{Name: "option3"},
+			}, tu.FieldName(multiSelectFieldName)),
```
ludydoo, line 49: I think these cases/tests should be kept!

PR #326, opened 2021-12-14
Go learning

neb42:
DO NOT MERGE

This is a bit of learning that would be good to get feedback on.

I've created a resource `Foo` and added the create flow (along with some extra store methods).

One thing that I've done differently is introducing the `FooReads` struct. This struct has no id field and uses pointers so that we can check for missing fields rather than zero values.

### DIFF
--- a/pkg/api/types/foo.go
+++ b/pkg/api/types/foo.go
@@ -1,0 +1,16 @@
+package types
+
+import uuid "github.com/satori/go.uuid"
+
+type FooReads struct {
+	Name       *string `json:"name"`
+	OtherField *int    `json:"otherField"`
+	Valid      *bool   `json:"valid"`
+}
+
+type Foo struct {
+	ID         uuid.UUID `json:"id"`
+	Name       string    `json:"name"`
+	OtherField int       `json:"otherField"`
+	Valid      bool      `json:"valid"`
+}
--- a/pkg/api/types/foo.go
+++ b/pkg/api/types/foo.go
@@ -1,16 +1,20 @@
 package types
 
-import uuid "github.com/satori/go.uuid"
+import (
+	uuid "github.com/satori/go.uuid"
+)
 
 type FooReads struct {
-	Name       *string `json:"name"`
-	OtherField *int    `json:"otherField"`
-	Valid      *bool   `json:"valid"`
+	Name       *string    `json:"name"`
+	OtherField *int       `json:"otherField"`
+	UUIDField  *uuid.UUID `json:"uuidField"`
+	Valid      *bool      `json:"valid"`
 }
 
 type Foo struct {
 	ID         uuid.UUID `json:"id"`
 	Name       string    `json:"name"`
 	OtherField int       `json:"otherField"`
+	UUIDField  uuid.UUID `json:"uuidField"`
 	Valid      bool      `json:"valid"`
 }
--- a/pkg/server/formsapi/handlers/foo/create.go
+++ b/pkg/server/formsapi/handlers/foo/create.go
@@ -1,0 +1,61 @@
+package foo
+
+import (
+	"net/http"
+
+	"github.com/emicklei/go-restful/v3"
+	"github.com/nrc-no/core/pkg/api/meta"
+	"github.com/nrc-no/core/pkg/api/types"
+	"github.com/nrc-no/core/pkg/api/types/validation"
+	"github.com/nrc-no/core/pkg/logging"
+	"github.com/nrc-no/core/pkg/utils"
+	uuid "github.com/satori/go.uuid"
+	"go.uber.org/zap"
+)
+
+func (h *Handler) Create() http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		ctx := req.Context()
+		l := logging.NewLogger(ctx)
+
+		var reads types.FooReads
+		if err := utils.BindJSON(req, &reads); err != nil {
+			l.Error("Failed to bind request body", zap.Error(err))
+			utils.ErrorResponse(w, err)
+			return
+		}
+
+		if errList := validation.ValidateFooReads(&reads); !errList.IsEmpty() {
+			err := meta.NewInvalid(types.FooGR, "", errList)
+			l.Error("Failed to validate request body", zap.Error(err))
+			utils.ErrorResponse(w, err)
+			return
+		}
+
+		// I don't like how there's no validation that each field is set.
+		// I get that they will default to a zero value
+		// But that isn't the same as forgetting to set it.
+		// Eg. here the validation will pass if I forget to set "valid"
+		// because it will default to false.
+		foo := types.Foo{
+			ID:         uuid.NewV4(),
+			Name:       *reads.Name,
+			OtherField: *reads.OtherField,
+			Valid:      *reads.Valid,
+		}
+
+		createdFoo, err := h.store.Create(ctx, &foo)
+		if err != nil {
+			l.Error("Failed to create foo", zap.Error(err))
+			utils.ErrorResponse(w, err)
+			return
+		}
+
+		utils.JSONResponse(w, http.StatusCreated, createdFoo)
+	}
+}
+
+func (h *Handler) RestfulCreate(request *restful.Request, response *restful.Response) {
+	handler := h.Create()
+	handler(response.ResponseWriter, request.Request)
+}
--- a/pkg/server/formsapi/handlers/foo/create.go
+++ b/pkg/server/formsapi/handlers/foo/create.go
@@ -13,38 +13,65 @@ import (
 	"go.uber.org/zap"
 )
 
+func ReadFoo(req *http.Request, l *zap.Logger) (*types.FooReads, error) {
+	var foo types.FooReads
+	if err := utils.BindJSON(req, &foo); err != nil {
+		l.Error("Failed to bind request body", zap.Error(err))
+		return nil, err
+	}
+
+	if errList := validation.ValidateFooReads(&foo); !errList.IsEmpty() {
+		err := meta.NewInvalid(types.FooGR, "", errList)
+		l.Error("Failed to validate request body", zap.Error(err))
+		return nil, err
+	}
+	return &foo, nil
+}
+
+func NewFoo(id uuid.UUID, name string, otherField int, uuidField uuid.UUID, valid bool, l *zap.Logger) (*types.Foo, error) {
+	foo := types.Foo{
+		ID:         id,
+		Name:       name,
+		OtherField: otherField,
+		UUIDField:  uuidField,
+		Valid:      valid,
+	}
+
+	// This is probably excessive
+	if errList := validation.ValidateFoo(&foo); !errList.IsEmpty() {
+		err := meta.NewInvalid(types.FooGR, "", errList)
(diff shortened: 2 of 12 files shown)


pkg/store/foo.go
```diff
@@ -0,0 +1,102 @@
...
+	"context"
+	"errors"
+
+	"github.com/nrc-no/core/pkg/api/types"
+	uuid "github.com/satori/go.uuid"
+	"go.uber.org/zap"
+	"gorm.io/gorm"
+)
+
+// Why do I have to define this again for the migrations?
```
ludydoo, line 13: You could, but then you'd have to add the `gorm` annotations on your api types, which is not so nice. Also, we have a some types that cannot be directly serialized to a flat hierarchy (such as form definitions) - so we need a different storage model

pkg/store/foo.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (d *fooStore) ListValid(ctx context.Context) ([]*types.Foo, error) {
+	_, db, l, done, err := actionContext(ctx, d.db, "foo", "list_valid")
+	if err != nil {
+		return nil, err
+	}
+	defer done()
+
+	var f []*types.Foo
+	if err := db.Where("valid IS TRUE").Find(&f).Error; err != nil {
```
ludydoo, line 65: that works, but I perfer `db.Where("valid = ?", true)` just so I use parametrized queries as much as possible, but in this case it doesn't really matter

pkg/server/public/handlers/foo/create.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+
+		if errList := validation.ValidateFooReads(&reads); !errList.IsEmpty() {
+			err := meta.NewInvalid(types.FooGR, "", errList)
+			l.Error("Failed to validate request body", zap.Error(err))
+			utils.ErrorResponse(w, err)
+			return
+		}
+
+		// I don't like how there's no validation that each field is set.
```
ludydoo, line 35: I usually name the fields with the fact that the zero value is the default, like
`enableDangerousLogs (default = false)`, `disableTls (default = false)`

If you really need the user to provide a value, you could set up a kind of enum (with `iota` + go genereate & `stringer` package) so that you enforce a value there.

But you need to keep in mind that this does not just apply to deserialization. You can pass any struct to any method and not put in some fields. It's the way it was designed.. 

Some people create mandatory constructors + do not expose the structs outside of the package sometimes. 

I think that if the validation package is well tested, and it catches the errors, then there is no real concern here..

pkg/api/types/foo.go
```diff
@@ -0,0 +1,16 @@
...
+
+import uuid "github.com/satori/go.uuid"
+
+type FooReads struct {
+	Name       *string `json:"name"`
+	OtherField *int    `json:"otherField"`
+	Valid      *bool   `json:"valid"`
+}
+
+type Foo struct {
```
ludydoo, line 11: If you want to have two types for this, you could embed FooReads into Foo
```
type Foo struct {
  FooReads
  ID uuid.UUID
}
```
then you can reuse the same validation logic for both types

pkg/store/foo.go
neb42: Makes sense. I guess if my Foo (api type) and Foo (db type) differed at all then it would throw an error when trying to write to the database?

pkg/server/public/handlers/foo/create.go
neb42: > But you need to keep in mind that this does not just apply to deserialization
Agreed, but I found it easier to cause some significant issues when it's not validated for deserialisation.

> Some people create mandatory constructors + do not expose the structs outside of the package sometimes.
Yes, I was thinking of doing something like this. I may give it a go.

neb42: Although actually I'm not sure how the constructor pattern would work:
```
# types/foo.go
package types

type foo struct {
  ...
}

func NewFoo( ... ) *foo {
  return &foo{ ... }
}
```

```
# store/foo.go

import "github.com/nrc-no/core/pkg/api/types"

# How do we get types.foo without exposing the struct?
func (d *fooStore) Create(ctx context.Context, foo *types.foo) (*types.foo, error) {
  ...
}
```

ludydoo: You can't pass non-exported types across packages...

neb42: Yeah exactly, so not sure this really works

ludydoo: I think it's too much investment for what it gives back... One thing you could do is to implement the `UnmarshalJSON` interface on your type, and verify that the types are there.
```
func (m *MyType) UnmarshalJSON(b []byte) error {
temp := map[string]interface{}
if err := json.Unmarshal(b, &temp); err != nil {
return err
}
if _, ok := temp["myJsonField"]; !ok {
  return fmt.ErrFmt("missing myJsonField")
}
return json.Marshal(b, &m)

}
```

pkg/store/foo.go
neb42: Yeah I'm not against raw sql. orms are good for simple things (like what I'm doing here), but for complex structures they can get a bit weird.

pkg/api/types/foo.go
gorm: @neb42 There might be good reasons using a pointer, but I would only use it if you have a good reason, e.g.need to differentiate between nil or default value. It's usually best to just use the default value and let the consuner handle it, but it might feel a bit strange in JS.

Alternatively, you could use the omitempty tag which would remove the property if it's set to a default value then use pointers if you can't differentiate e.g. on bool or a nested struct.

https://www.sohamkamani.com/golang/omitempty/

Might not be an issue for you, but using pointers will also lead to more memory usage, need to store both the pointer and the value, slower json encoding and it also leads to more complexity. So in general it should be avoided unless it's needed as it's seldom an optimization.

The same applies to small struct passed as arguments
https://github.com/golang/go/wiki/CodeReviewComments#pass-values.

Here are aom good rules for receivers and pointers https://github.com/golang/go/wiki/CodeReviewComments#pass-values

Go is fast at copying data so e.g. returning a pointer from a function is also something you should have good reasons to do as it's not really an optimization (lead to more complex code and data would probably end up on the heap and require GC instead of on the stack).

If you have a choice, try to optimize for simplicity and maintainability. Clear is better than clever.

neb42: > If you have a choice, try to optimize for simplicity and maintainability. Clear is better than clever

100% agree and thanks for the links, I'll definitely give them a read through.

I think my issue is that default values lead to bad api design (maybe slightly too strong a claim). On the client side I find it strange that not sending a field will give it a default value rather than an error. I personally prefer a little more internal complexity for a cleaner interface.

I guess this is just solved with good validation rules that error if the default value is invalid, eg. a name can't be an empty string.

I think it also introduces an extra layer of complexity when a value could be null.

gorm: The philosophy between Go and JavaScript is a bit different here. In old days we would just check for falsy in JS as it would also include the default values of false, "" and 0. Now I guess it's cleaner to use omitempty tag on the Go side and then just check for foo === undefined or foo ?? error();

gorm: Regarding pointers, this is also an interesting article https://medium.com/@vCabbage/go-are-pointers-a-performance-optimization-a95840d3ef85

neb42: Haha, I think javascript also has it's issues with falsy values.

I think checking for undefined on the response is fine (not my preference but I'm ok with it)

My issue is more on the request. I think if a field is missing it should return a 422 rather than give a default value.

gorm: Aha I didn't understand this was for a request. Personally I would just check for empty value.  Or use a pointer like you did. A few other options as well https://stackoverflow.com/questions/19633763/unmarshaling-json-in-go-required-field

neb42: Yeah I think my approach is to define a `reads` struct using pointers, allowing us to check for `nil` where appropriate. This then gets put into an internal struct (which I also use for the `writes`) using values.

Although if I refactored it to use:
```
type Foo struct {
  FooReads
  ID uuid.UUID
}
```
then this would still use pointers.

> Personally I would just check for empty value

I think this would mostly work, until it doesn't. For strings, I would typically have an empty string and null be the same so I can live with that. Booleans are fine as false. It's numbers having the default as 0 that causes an issue for me. Javascript also annoys me with 0's being falsy.

gorm: Yea. SQL has nullable types, but you need to handle that yourself in JSON, or you could use https://github.com/guregu/null

neb42: Yeah I did see that package a while ago. It may be worth trying out if this becomes an issue. I think I need to spend more time with go first though (my backend background is with node/scala/python).

pkg/api/types/foo.go
```diff
@@ -0,0 +1,16 @@
...
+	ID         uuid.UUID `json:"id"`
```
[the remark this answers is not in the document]
neb42: Haven't checked the openapi docs, but it seems to marshal fine. I'll try adding an extra field to test the unmarshal.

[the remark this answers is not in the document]
neb42:
```
code: 400
message: "failed to unmarshal json body: uuid: incorrect UUID length: dsfsdfsdfsdf"
reason: "BadRequest"
status: "Failure"
```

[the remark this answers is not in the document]
ludydoo: can you check what the `localhost:9000/openapi.json` document returns ?

pkg/server/formsapi/handlers/foo/create.go
```diff
@@ -13,38 +13,65 @@ import (
 	"go.uber.org/zap"
 )
 
+func ReadFoo(req *http.Request, l *zap.Logger) (*types.FooReads, error) {
```
ludydoo, line 16: If you have optional fields, you could use a pattern quite commonly used. Basically definition "option functions" for your `Foo` and passing them as a vararg parameter.

```
type FooOption func(foo *types.FooReads) *types.FooReads
var ValidFoo(isValid bool) FooOption = func(foo *types.FooReads) *types.FooReads {
  foo.valid = isValid
  return foo
}

func NewFoo(id uuid.UID, name string, options ...FooOption) *types.FooReads{
  foo := &types.FooReads{ name: name}
  for _, option := range options {
    foo = option(foo)
  }
  return foo
}

# otherpkg

var foo = NewFoo(uuid.NewV4(), "name", types.ValidFoo(false))

```

But you still have the problem that if a package function expects a `Foo`, nothing prevents the user from bypassing your constructor altogether.
```
package mypkg

function StoreFoo(foo *types.Foo) {...}

package myotherpkg

...
mypkg.StoreFoo(&types.Foo{}) // <- bypassed constructor
```

I think the only way you could really enforce this would be to use interfaces and not export the struct...

```
type Foo interface {
  GetName() string
  GetValid() bool
}

type foo struct{
 name string
 valid bool
}

func (f foo) GetName() string { return f.name}

func (f foo) GetValid() string { return f.valid}

func NewFoo(name string, valid string) Foo {
  return &foo{name: name, valid: valid}
}

```

PR #327, opened 2021-12-15
DS-41: theming

internetti:


neb42:
It's weird some of these linting errors didn't get fixed when I ran `yarn lint:fix`

frontend/packages/design-system/src/themes/Button.ts
```diff
@@ -0,0 +1,74 @@
+export default {
+  variants: {
+    major: ({ colorScheme }: any) => {
```
neb42, line 3: I think `major` and `minor` aren't descriptive enough. My choice would be `filled` and `outlined`.

internetti: maxime has been saying he prefers to name stuff after functionality, not looks, in case something is redesigned. e.g. if they decide to signify the lower significance of a button with something else than an outline, we have to rename this.
don't know how likely that is at this point, but that reasoning made sense to me. what do you think about that?

neb42: Yeah I understand his reasoning and agree with it, but I'm not sure it should be applied everywhere. When defining colours, spacing, text, it makes sense. But if you look at any other design system that separates colour and style, it uses something descriptive for the style. [IBM](https://www.carbondesignsystem.com/components/button/usage/#live-demo) is an example that doesn't do this, as their button style is tied to the colour.

I think if we do something like this we need a more extensible naming convention, like `a, b, c`, but no great ideas off the top of my head.

It may be worth merging this pr for now seen as it's not really in use and we can bike shed on naming afterwards.

ludydoo: @internetti  @neb42  lgtm? I don't have any opinion on this

PR #333, opened 2021-12-16
Rename admin frontend to core-authnz-frontend

ludydoo:


frontend/apps/core-authnz-frontend/package.json
```diff
@@ -1,9 +1,9 @@
 {
-  "name": "admin",
+  "name": "@nrc.no/core-authnz-frontend",
```
neb42, line 2: I've been meaning to ask, is this the scope we want to be using? I'll reserve the npm org if so.

PR #341, opened 2021-12-17
DS-43: spacing

internetti:


frontend/packages/design-system/src/themes/index.ts
```diff
@@ -9,4 +9,13 @@ export default extendTheme({
   components: {
     Button,
   },
+  space: {
+    '1': tokens.spacing['spacing-5'],
```
neb42, line 13: I have a small preference for `spacing5` so that we can do `tokens.spacing.spacing5`

PR #343, opened 2021-12-20
COR-445-Colombia_Seeder--Seeder_Skeleton

nilueps:
Note: following @neb42 's useful advice, this PR is not a request to merge into main, but rather into the COR-445 branch.

with the server running in the background, invoke the script with 
`go run . seed -e "https://localhost:9000" global`

nb: these scripts aren't idempotent, if you need to clear the db you can do so with the following:
`VERY_DANGEROUSLY_ALLOW_TO_CLEAR_DB=yes make clear-db`


internetti:
could you add some info on how to run this? or add some tests?

nilueps:
Yea I admit it's a little obscure; I will do that ;)

PR #346, opened 2021-12-21
Core app setup

neb42:
* Sets up the skeleton structure for the native app
* Added the auth flow code and the root navigation
* Slight refactor to the auth code to make it easier to test
* Auth flow tests
* Left old native code in place as we may want to pull bits of it in as we build out features
* Change linting line length to 80
* In serve-backend check that pids are running before stopping (sometimes the tunnels would timeout and close themselves)


frontend/.prettierrc.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  printWidth: 130,
+  printWidth: 80,
```
nilueps, line 2: do we really have such small screens? ^^

neb42: @internetti uses multiple panes alongside one another. I think 80 is also the recommended setting in python linting based on that. Typically I go with 100.

frontend/apps/core-app/src/components/AuthWrapper/useTokenResponse.ts
```diff
@@ -0,0 +1,101 @@
...
+        const tr = await exchangeCodeAsync(exchangeConfig, discovery);
+        setTokenResponse(tr);
+      } catch {
+        setTokenResponse(undefined);
+      }
+    })();
+  }, [
+    request?.codeVerifier,
+    JSON.stringify(response),
+    JSON.stringify(discovery),
```
nilueps, line 69: Just out of curiosity, why are you stringifying here? Is it because `useEffect` doesn't see changes to deeply nested object properties?

frontend/apps/core-app/src/components/AuthWrapper/useTokenResponse.ts
```diff
@@ -0,0 +1,101 @@
...
+          extraParams: {},
+        };
+
+        try {
+          const resp = await tokenResponse?.refreshAsync(
+            refreshConfig,
+            discovery,
+          );
+          setTokenResponse(resp);
+        } catch {
```
nilueps, line 89: Is ignoring the exception voluntary?

neb42: You mean instead of:
```
} catch (err) {
```

Yes, it wasn't always but was added 2/3 years ago I think.

PR #350, opened 2021-12-21
COR-445-Colombia Seeder: fix seeding error

nilueps:
Our PR wasn't exhaustive enough and an error slipped through.

### DIFF
--- a/pkg/seeder/seed_colombia.go
+++ b/pkg/seeder/seed_colombia.go
@@ -18,6 +18,19 @@ func (s *Seed) seedColombia(ctx context.Context, client client.Client) error {
 		return err
 	}
 
+	// keep a reference to the global beneficiary form as a reference field for use in the forms
+	s.globalBeneficiaryRefField = &types.FieldDefinition{
+		Name:     "Individual Beneficiary",
+		Key:      true,
+		Required: true,
+		FieldType: types.FieldType{
+			Reference: &types.FieldTypeReference{
+				DatabaseID: s.globalDatabase.ID,
+				FormID:     s.globalRootBeneficiaryForm.ID,
+			},
+		},
+	}
+
 	err = s.seedCoIntake(ctx, client, co_db.ID)
 
 	if err != nil {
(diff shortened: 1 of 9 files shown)


pkg/seeder/seed_colombia.go
```diff
@@ -45,19 +58,10 @@ func (s *Seed) seedCoIntake(ctx context.Context, client client.Client, dbID stri
...
-		intake_contact_info,
-		intake_ind_cc_specific,
-		intake_hh_information,
-	}
-
 	err = client.CreateFolder(ctx, &types.Folder{
 		Name:       "Intake",
 		DatabaseID: dbID,
-	}, intake_folder)
+	}, &intake_folder)
```
internetti, line 64: there's a test in `pkg/server/formsapi/tests/helpers_test.go` that still uses the old version of this change

internetti (dismissed):

pkg/seeder/seed_colombia.go
nilueps: I'm not sure what you're referring to, looks like all tests passed on prow, and I just double-checked on local too

internetti: when I check out your branch, there is this helper tests file that has this function
`func (s *Suite) createFolder(ctx context.Context, dbId string, out *types.Folder) error {
	if err := s.cli.CreateFolder(ctx, &types.Folder{
		Name:       "My Folder",
		DatabaseID: dbId,
	}, out); err != nil {
		return err
	}
	return nil
}
`
this one still passes a folder pointer to the function. I don't know why this is not breaking the tests, maybe this test is not being run?

internetti: this function doesn't seem to have usages, maybe you could delete it?

nilueps: I mean, this change is a noop anyway. I think I introduced it because I was having some problems and I tried this just in case. A pointer is being passed in both cases, if you look a few lines above you'll see the type changed from `*types.Folder` to `types.Folder`

PR #351, opened 2021-12-22
DS-33: buttons

internetti:


frontend/packages/design-system/src/assets/icons/ActivityCamp.tsx
```diff
@@ -1,13 +1,11 @@
...
-
-export default (variant: IconVariants) => {
-  return (
-    <Path
-      fill={variant}
-      d="M12,18.2l8-4.6l8,4.6c0.2,0.1,0.4,0.2,0.6,0.2c0.4,0,0.8-0.2,1.1-0.6c0.3-0.6,0.1-1.3-0.5-1.7l-8.6-4.9
+export const ActivityCamp = (
+  <Path
+    fill="currentColor"
+    stroke="currentColor"
```
neb42, line 7: I don't think we want to set `fill` and `stroke`. When the designer makes an icon they will set which property uses the colour. We should just replace that colour with `currentColor`. Here it would be `fill`.

frontend/packages/design-system/src/assets/icons/ActivityCamp.tsx
```diff
@@ -1,13 +1,11 @@
...
 import { Path } from 'react-native-svg';
 
-import { IconVariants } from '../../types/icons';
-
-export default (variant: IconVariants) => {
-  return (
-    <Path
-      fill={variant}
-      d="M12,18.2l8-4.6l8,4.6c0.2,0.1,0.4,0.2,0.6,0.2c0.4,0,0.8-0.2,1.1-0.6c0.3-0.6,0.1-1.3-0.5-1.7l-8.6-4.9
+export const ActivityCamp = (
```
neb42, line 4:
```
export const ActivityCamp = () => (
 <Path ... />
);
```

internetti: that caused typescript issues in Icon.tsx

frontend/packages/design-system/src/components/Icon.tsx
```diff
@@ -1,163 +1,31 @@
...
-      };
-  }
+const Icon: FC<Props> = (props) => {
+  const { name } = props;
+  const MappedIcon = iconMap[name];
 
   return (
-    <IconNB viewBox="0 0 40 40" style={{ height: '40px', width: '40px', display: 'flex' }} children={iconComponent(variant)} />
+    <IconNB
+      viewBox="0 0 40 40"
```
neb42, line 17: I don't think we want to be setting any props here. These should all be set through native base.

internetti: but we do want a fix size, don't we?

frontend/packages/design-system/src/components/Icon.tsx
```diff
@@ -1,163 +1,31 @@
...
-      break;
-    case IconName.ACTIVITY_WASH:
-      iconComponent = icons.ActivityWash;
-      break;
-    default:
-      iconComponent = () => {
-        return <Path />;
-      };
-  }
+const Icon: FC<Props> = (props) => {
```
neb42, line 11: Props can be destructed here. I don't think we want to include `name` in the props we send to the native base icon.

`const Icon: FC<Props> = ({ name, ...props }) => {`

frontend/packages/design-system/src/components/Icon.tsx
```diff
@@ -1,163 +1,31 @@
...
+  const { name } = props;
+  const MappedIcon = iconMap[name];
```
neb42, line 13: I think a default icon would be good here to prevent errors

frontend/packages/design-system/src/components/Icon.tsx
internetti: got an error about rest properties from typescript that way

frontend/packages/design-system/src/components/Icon.tsx
```diff
@@ -1,163 +1,31 @@
...
+import { Icon as IconNB, IIconProps } from 'native-base';
 
-import { IconName, IconProps } from '../types/icons';
+import { iconMap } from '../assets/iconMap';
+import { IconName } from '../types/icons';
 
-import * as icons from './icons';
+type Props = {
+  name: keyof typeof iconMap;
+} & Omit<IIconProps, 'name'> & { name: IconName };
```
neb42, line 9: Why do we have `& { name: IconName }`? This is overriding `name: keyof typeof iconMap;`

frontend/packages/design-system/src/components/Icon.tsx
internetti: native base seems to put a help icon automatically

frontend/packages/design-system/src/assets/icons/ActivityCamp.tsx
neb42: What was the error? I'm not seeing anything

frontend/packages/design-system/src/components/Icon.tsx
neb42: I'm also not getting an error for this. What was it?

frontend/packages/design-system/src/components/Icon.tsx
neb42: ah nice, although if we changed it to be this:
```
const MappedIcon  = iconMap[name];

  return (
    <IconNB {...props}>
      <MappedIcon />
    </IconNB>
  );
```
We probable still need some handling for when it's `undefined`

frontend/packages/design-system/src/components/Icon.tsx
neb42: I don't think so. In my experience icon sizes vary. If it is this way in the designs now I don't think we should restrict ourselves to it. If we think there's a fixed size we want I would set the `defaultProps` for `Icon` in the theme.

frontend/packages/design-system/src/types/icons.ts
```diff
@@ -1,14 +1,8 @@
...
+  color: string;
 };
 
-export enum IconVariants {
-  DARK = '#000000',
-  LIGHT = '#FFFFFF',
-  LINK = '#1E4A7D',
-}
-
 export enum IconName {
```
neb42, line 6: Do we need this `enum`? If the `name` prop is `keyof typeof iconMap` then I'm not sure this is necessary and it is difficult to mantain.

frontend/packages/design-system/src/types/icons.ts
```diff
@@ -1,14 +1,8 @@
 export type IconProps = {
```
neb42, line 1: Is this still used?

frontend/packages/design-system/src/components/Icon.tsx
internetti: when MappedIcon is undefined, native base shows a help icon. problem is just that that one is for some reason not aligned correctly

frontend/packages/design-system/src/types/icons.ts
internetti: yeah, this is currently only used for the add ons in storybook, might be replaceable

frontend/packages/design-system/src/types/icons.ts
internetti: this one not, I'll move the updated type here though

frontend/packages/design-system/src/types/icons.ts
internetti: but then we would have to export iconMap, not sure if we want that

frontend/packages/design-system/src/components/Icon.tsx
neb42: I mean if `MappedIcon` is undefined then `<MappedIcon />` will error. It should probably be `{MappedIcon && <MappedIcon />}`. Then if it's undefined the default native base icon will show.

neb42: I don't think it's that bad if it doesn't align, as we don't want to be in that situation anyway

frontend/packages/design-system/src/types/icons.ts
neb42: Yeah I'd probably replace it with something like `keyof typeof iconMap`. That way we don't have to remember to add it here when adding new icons.

frontend/packages/design-system/src/assets/icons/ActivityCamp.tsx
internetti: it was something about this not being callable. now it's of course not happening anymore

frontend/packages/design-system/src/components/Icon.tsx
internetti: until the last change, it was {MappedIcon}, so it worked, but with the icons being functions again, I'll have to change it to that yes

frontend/packages/design-system/src/components/Icon.tsx
internetti: that's how I understood the designers, but okay, then I'll set some defaults

frontend/packages/design-system/src/types/icons.ts
```diff
@@ -1,60 +1,7 @@
...
-  USER_GROUP_ADD = 'user_group_add',
-  WHATSAPP = 'whatsapp',
-  ACTIVITY_CAMP = 'activity_camp',
-  ACTIVITY_EDUCATION = 'activity_education',
-  ACTIVITY_FOOD = 'activity_food',
-  ACTIVITY_ICLA = 'activity_icla',
-  ACTIVITY_SHELTER = 'activity_shelter',
-  ACTIVITY_WASH = 'activity_wash',
-}
+export type IconProps = {
```
neb42, line 5: I generally think props are best left next to their component, unless there's a very strong reason for sharing them?

frontend/storybook/storybook/stories/Icon/Icon.stories.js
```diff
@@ -1,21 +1,20 @@
 import { storiesOf } from '@storybook/react-native';
 import React from 'react';
-import { Icon } from 'core-design-system';
-import { IconName, IconVariants } from 'core-design-system/lib/esm/types/icons';
+import { Icon, icons, theme, tokens } from 'core-design-system';
+import { IconName } from 'core-design-system/lib/esm/types/icons';
```
neb42, line 4: Does `IconName` need removing here?

frontend/storybook/storybook/stories/Icon/Icon.stories.js
```diff
@@ -1,21 +1,20 @@
 import { storiesOf } from '@storybook/react-native';
 import React from 'react';
-import { Icon } from 'core-design-system';
-import { IconName, IconVariants } from 'core-design-system/lib/esm/types/icons';
+import { Icon, icons, theme, tokens } from 'core-design-system';
```
neb42, line 3: If you don't want to have `icons` as an export of the package, you could do:
```
import icons from 'core-design-system/lib/esm/cant-remember-the-path';
```

But I think the export is fine

frontend/packages/design-system/src/types/icons.ts
internetti: this is a small project (so far), but in general it happens a lot that types are used by multiple components or other types, and then there's the question of where to put them, and I don't like having some types with the components and some types somewhere else, so I like to have a types folder with all the types

frontend/packages/design-system/src/themes/Button.ts
```diff
@@ -3,14 +3,16 @@ export default {
...
         backgroundColor: `${colorScheme}.500`,
+
         _text: {
           color: 'white',
           bold: true,
         },
         _hover: {
-          backgroundColor: `${colorScheme}.100`,
+          backgroundColor:
+            colorScheme === 'secondary' ? 'link' : `${colorScheme}.200`,
```
neb42, line 13: Not sure what you or the designers think about this, but I feel this should be `secondary.xxx` as the intention of this component isn't a "link" it's a button. I'm not sure the style of the button should be tied to the style of a link.

PR #352, opened 2021-12-22
COR-446-Fix_PWA_forms_display

nilueps:


frontend/apps/pwa/src/reducers/form.ts
```diff
@@ -252,13 +290,28 @@ export const selectFormOrSubFormById = (state: RootState, formOrSubFormId: strin
   return undefined;
 };
 
-export const selectByFolderOrDBId = (state: RootState, folderOrDbId?: string): FormDefinition[] => {
+export const selectByFolderOrDBId = (
```
internetti, line 293: could you add some tests for this one?

frontend/apps/pwa/src/features/browser/FolderBrowser.tsx
```diff
@@ -123,7 +123,7 @@ export const FolderBrowserContainer: FC<FolderBrowserContainerProps> = (props) =
   }, [dispatch]);
 
   const childFolders = useAppSelector(selectChildFolders(props.folderId ? props.folderId : props.databaseId));
-  const childForms = useAppSelector(selectChildForms(props.folderId ? props.folderId : props.databaseId));
+  const childForms = useAppSelector(selectChildForms({dbId: props.databaseId, folderId: props.folderId}));
```
neb42, line 126: It looks like the lint commit hook isn't working. I think it'd also be good to setup lint on save in your editor. You use vim right? https://medium.com/@jimeno0/eslint-and-prettier-in-vim-neovim-7e45f85cf8f9

nilueps: thanks for the link, @neb42 

I've been calling `eslint --fix` manually so far, which is easy to forget 😅 

For the git hook, it may be that husky doesn't work from a subdirectory? https://scottsauber.com/2021/06/01/using-husky-git-hooks-and-lint-staged-with-nested-folders/

Perhaps setting up `frontend` as a submodule could work? https://git-scm.com/book/en/v2/Git-Tools-Submodules

PR #353, opened 2021-12-22
COR-425-Fix_token_renewal

nilueps:
This probably isn't the most elegant solution, but I left the browser window open for over 30 minutes and I wasn't automatically logged out as was the case previously.

Edit: the refresh gets triggered, but I'm still getting an unauthorized response

### DIFF
--- a/frontend/apps/core-app/src/components/AuthWrapper/useTokenResponse.ts
+++ b/frontend/apps/core-app/src/components/AuthWrapper/useTokenResponse.ts
@@ -69,29 +69,44 @@ export const useTokenResponse = (): [
     JSON.stringify(discovery),
   ]);
 
+  const refreshTokenInterval = React.useRef<number | null>(null);
   React.useEffect(() => {
-    (async () => {
+    const refreshToken = async () => {
       if (!discovery) return;
+      if (!tokenResponse) return;
+      if (!tokenResponse.shouldRefresh()) return;
 
-      if (tokenResponse?.shouldRefresh()) {
-        const refreshConfig = {
-          clientId,
-          scopes: Constants.manifest?.extra?.scopes,
-          extraParams: {},
-        };
+      const refreshConfig = {
+        clientId,
+        scopes: Constants.manifest?.extra?.scopes,
+        extraParams: {},
+      };
 
-        try {
-          const resp = await tokenResponse?.refreshAsync(
-            refreshConfig,
-            discovery,
-          );
-          setTokenResponse(resp);
-        } catch {
-          setTokenResponse(undefined);
-        }
+      try {
+        const resp = await tokenResponse.refreshAsync(refreshConfig, discovery);
+        setTokenResponse(resp);
+      } catch (err) {
+        setTokenResponse(undefined);
+        throw err;
       }
-    })();
-  }, [tokenResponse?.shouldRefresh(), discovery]);
+    };
+
+    if (refreshTokenInterval.current)
+      window.clearInterval(refreshTokenInterval.current);
+
+    if (
+      tokenResponse &&
+      tokenResponse.expiresIn &&
+      tokenResponse.expiresIn > 0
+    ) {
+      refreshTokenInterval.current = window.setInterval(refreshToken, 1000);
+    }
+
+    return () => {
+      if (refreshTokenInterval.current)
+        clearInterval(refreshTokenInterval.current);
+    };
+  }, [tokenResponse?.refreshToken, tokenResponse?.expiresIn]);
 
   const login = React.useCallback(async () => {
     promptAsync({ useProxy: shouldUseProxy });
--- a/frontend/packages/core-api-client/src/lib/BaseRESTClient.ts
+++ b/frontend/packages/core-api-client/src/lib/BaseRESTClient.ts
@@ -8,8 +8,6 @@ export class BaseRESTClient {
 
   private token = '';
 
-  lastInterceptorId: number | undefined;
-
   constructor(baseURL: string) {
     this.axiosInstance = axios.create({
       baseURL,
@@ -27,11 +25,6 @@ export class BaseRESTClient {
       };
     });
   }
-
-  // this method became redundant, could be removed
-  public setAuth = (token: string): void => {
-    this.setToken(token);
-  };
 
   public getToken = (): string => this.token;
 
--- a/frontend/packages/core-api-client/src/lib/BaseRESTClient.ts
+++ b/frontend/packages/core-api-client/src/lib/BaseRESTClient.ts
@@ -6,42 +6,44 @@ import { clientResponse } from './utils/responses';
 export class BaseRESTClient {
   protected readonly axiosInstance: AxiosInstance;
 
+  private token = '';
+
   lastInterceptorId: number | undefined;
 
   constructor(baseURL: string) {
     this.axiosInstance = axios.create({
       baseURL,
     });
+    this.axiosInstance.interceptors.request.use((value: AxiosRequestConfig) => {
+      if (!this.getToken()) {
+        return value;
+      }
+      return {
+        ...value,
+        headers: {
+          ...value.headers,
+          Authorization: `Bearer ${this.getToken()}`,
+        },
+      };
+    });
   }
 
+  // this method became redundant, could be removed
   public setAuth = (token: string): void => {
-    console.log('Setting token to:', token);
+    this.setToken(token);
+  };
 
-    if (this.lastInterceptorId != null)
-      this.axiosInstance.interceptors.request.eject(this.lastInterceptorId);
+  public getToken = (): string => this.token;
 
-    this.lastInterceptorId = this.axiosInstance.interceptors.request.use(
-      (value: AxiosRequestConfig) => {
-        if (!token) {
-          return value;
-        }
-        const headers = {
-          ...value.headers,
-          Authorization: `Bearer ${token}`,
-        };
-        value.headers = headers;
-        console.log(value.headers.Authorization);
-        console.log(value.url);
-        return value;
-      },
-    );
+  public setToken = (token: string): void => {
+    this.token = token;
   };
 
   protected async do<TRequest, TBody>(
     request: TRequest,
     url: string,
     method: Method,
-    data: any,
+    data: unknown,
     expectStatusCode: number,
     options?: RequestOptions,
   ): Promise<Response<TRequest, TBody>> {
--- a/frontend/packages/core-api-client/src/lib/BaseRESTClient.ts
+++ b/frontend/packages/core-api-client/src/lib/BaseRESTClient.ts
@@ -1,10 +1,12 @@
-import axios, { AxiosInstance, Method } from 'axios';
+import axios, { AxiosInstance, AxiosRequestConfig, Method } from 'axios';
 
 import { RequestOptions, Response } from './types';
 import { clientResponse } from './utils/responses';
 
 export class BaseRESTClient {
   protected readonly axiosInstance: AxiosInstance;
+
+  lastInterceptorId: number | undefined;
 
   constructor(baseURL: string) {
     this.axiosInstance = axios.create({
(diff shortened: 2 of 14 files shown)


nilueps:
Don't merge this, I just realized that my fix isn't working for POST requests somehow

internetti (commented):
I think it would be better to have a look at the TokenRequest.isTokenFresh function, maybe that one is not working correctly. we could write a test for it and see if it behaves like we expect.

since we actually have tokens with expiration dates, I think we should use those expiration dates instead of a fix interval

frontend/packages/core-auth/src/components/AuthWrapper.tsx
```diff
@@ -98,17 +96,20 @@ const AuthWrapper: React.FC<AuthWrapperProps> = ({
...
           setTokenResponse(resp);
         })
-        .catch((err) => {
+        .catch(() => {
           setTokenResponse(undefined);
         });
     }
-  }, [tokenResponse?.shouldRefresh(), discovery]);
+  };
+  let refreshTokenIntervalId: NodeJS.Timeout | null = null;
```
neb42, line 104: I think you probably want this to be a `ref` so that is doesn't get set to null on each render.

```
const refreshTokenIntervalId = React.useRef<NodeJS.Timeout | null>(null);
...
if (refreshTokenIntervalId.current != null) clearInterval(refreshTokenIntervalId.current);
refreshTokenIntervalId.current = setInterval(()=>refreshToken(), 5000);
```

nilueps: That's neat, I didn't know about `useReft`, thanks

frontend/packages/core-auth/src/components/AuthWrapper.tsx
```diff
@@ -83,29 +80,29 @@ const AuthWrapper: React.FC<AuthWrapperProps> = ({
...
+        break;
+      case 'id_token':
+        token = tokenResponse?.idToken ?? '';
+        break;
+      default:
+        token = '';
     }
-  }, [tokenResponse?.shouldRefresh(), discovery]);
+    onTokenChange(token);
+  }, [JSON.stringify(tokenResponse)]);
```
nilueps, line 101: The functionality is nearly there, but for some reason that I don't understand, this `useEffect` hook is not getting triggered when the timeout triggers the token refresh.

The `setTokenResponse` call on line 160 doesn't cause a re-render; which is what I would [expect](https://stackoverflow.com/questions/46240647/react-how-to-force-a-function-component-to-render/53837442#53837442)...  I must be overlooking something

frontend/packages/core-api-client/src/lib/BaseRESTClient.ts
```diff
@@ -1,31 +1,40 @@
...
-      if (!token) {
-        return value;
-      }
-      return {
-        ...result,
-        headers: {
-          ...result.headers,
+    console.log('Setting token to:', token);
+
+    if (this.lastInterceptorId != null)
```
neb42, line 20: I think rather than ejecting the interceptor and using a new one, we could do a slight refactor. 

* Move the interceptor logic into the constructor
* Rather than taking `token` as a function argument, it's a method on the class `this.getToken()`. This will be called on each intercept.
* Add another class method `setToken` that will just store it as a property.
* When using `AuthWrapper` use `setToken` instead of `setAuth`

neb42: This package also seems useful, but I'm not sure it fits in with how we've structured auth. Potentially something for a refactor later down the line.
https://www.npmjs.com/package/axios-auth-refresh

neb42:
The refresh logic should probably handle a response returning a `401`, refreshing the token, and trying again.

frontend/packages/core-auth/src/components/AuthWrapper.tsx
neb42: `[JSON.stringify(tokenResponse)]` I don't think this will stringify that well. When using `JSON.stringify` it should be something json serialisable.

I'd replace this with `[tokenResponse?.accessToken, tokenResponse?.idToken]`

frontend/packages/core-auth/src/components/AuthWrapper.tsx
```diff
@@ -42,27 +51,15 @@ const AuthWrapper: React.FC<AuthWrapperProps> = ({
...
-    switch (injectToken) {
-      case 'access_token':
-        return tokenResponse?.accessToken ?? '';
-      case 'id_token':
-        return tokenResponse?.idToken ?? '';
-      default:
-        return '';
-    }
-  })();
-
```
neb42, line 55: Why did you remove this block?

frontend/packages/core-auth/src/components/AuthWrapper.tsx
```diff
@@ -114,16 +111,57 @@ const AuthWrapper: React.FC<AuthWrapperProps> = ({
...
+      if (delay <= 0) {
+        refreshToken();
+      } else {
+        refreshTokenTimeout.current = window.setTimeout(
+          () => refreshToken(),
+          delay,
+        );
+      }
+    }
+  }, [JSON.stringify(tokenResponse), discovery]);
```
neb42, line 137: Same here. I think `discovery` is ok, but you should check against the bits of `tokenResponse` you're interested in.

frontend/packages/core-auth/src/components/AuthWrapper.tsx
```diff
@@ -114,16 +111,57 @@ const AuthWrapper: React.FC<AuthWrapperProps> = ({
     }
   }, [tokenResponse, isLoggedIn]);
 
+  // Schedule a token refresh before it expires (or use an existing schedule)
+  const refreshTokenTimeout = useRef<number | null>(null);
   useEffect(() => {
-    onTokenChange(token);
-  }, [token]);
+    if (
+      tokenResponse != null &&
```
neb42, line 118: All of these should be `!==` unless there's a good reason

frontend/packages/core-auth/src/components/AuthWrapper.tsx
```diff
@@ -114,16 +111,57 @@ const AuthWrapper: React.FC<AuthWrapperProps> = ({
...
+    }
+  }, [JSON.stringify(tokenResponse), discovery]);
 
   const handleLogin = useCallback(() => {
     promptAsync().catch((err) => {
       handleLoginErr(err);
     });
   }, [discovery, request, promptAsync]);
 
+  async function refreshToken() {
```
neb42, line 145: This is defined after it is used. I think this is a linting error and generally makes the code harder to read.

frontend/packages/core-auth/src/components/AuthWrapper.tsx
```diff
@@ -114,16 +111,57 @@ const AuthWrapper: React.FC<AuthWrapperProps> = ({
...
+      console.log({ delay });
+
+      if (delay <= 0) {
+        refreshToken();
+      } else {
+        refreshTokenTimeout.current = window.setTimeout(
+          () => refreshToken(),
+          delay,
+        );
+      }
```
neb42, line 135: I feel this is going to end up with some failed requests if you make a request while refreshing. It may be better to refresh more frequently while the access token is still valid. Eg. do this on a `setInterval` at 1/3 of the expiration time and always refresh.

frontend/packages/core-auth/src/components/AuthWrapper.tsx
nilueps: I didn't, it just moved to line 89

frontend/packages/core-auth/src/components/AuthWrapper.tsx
nilueps: Switch to truthy checks for object and `== / !=` for primitives. https://basarat.gitbook.io/typescript/styleguide#null-vs.-undefined

frontend/packages/core-api-client/src/lib/BaseRESTClient.ts
```diff
@@ -1,38 +1,49 @@
...
 
 import { RequestOptions, Response } from './types';
 import { clientResponse } from './utils/responses';
 
 export class BaseRESTClient {
   protected readonly axiosInstance: AxiosInstance;
 
+  private token = '';
+
+  lastInterceptorId: number | undefined;
```
neb42, line 11: This isn't needed

frontend/packages/core-auth/src/types/response.ts
```diff
@@ -23,15 +28,18 @@ export class TokenResponse implements TokenResponseConfig {
   public constructor(response: TokenResponseConfig) {
     this.accessToken = response.accessToken;
     this.tokenType = response.tokenType ?? TokenType.Bearer;
-    this.expiresIn = response.expiresIn;
+    this.expiresIn = 10; // response.expiresIn;
```
neb42, line 31: I don't think this should have been committed?

frontend/packages/core-auth/src/types/response.ts
```diff
@@ -83,15 +92,30 @@ export class TokenResponse implements TokenResponseConfig {
...
+      response = await request.performAsync(discovery);
+    } catch (e) {
+      throw new Error(`Error encountered while performing token refresh: ${e}`);
+    }
+
     response.refreshToken = response.refreshToken ?? this.refreshToken;
     const json = response.getRequestConfig();
     this.applyResponseConfig(json);
+
     return this;
```
neb42, line 119: Why not this? The bit after the try/catch seems like it will never run on error and if it did then `response` could be `undefined` and we would error on `response.refreshToken = response.refreshToken ?? this.refreshToken;`. Putting it all in the try prevents this and also allows the response to be a `const` (I typically try to avoid using `let` unless a variable has a strong case for being reassigned).
```
try {
  const response = await request.performAsync(discovery);
  response.refreshToken = response.refreshToken ?? this.refreshToken;
  const json = response.getRequestConfig();
  this.applyResponseConfig(json);
  return this;
} catch (e) {
  throw new Error(`Error encountered while performing token refresh: ${e}`);
}
```

frontend/packages/core-auth/src/types/response.ts
```diff
@@ -83,15 +92,30 @@ export class TokenResponse implements TokenResponseConfig {
     return !(TokenResponse.isTokenFresh(this) || !this.refreshToken);
   }
 
+  getExpiryMs(): number {
+    const tenPercentMargin = 0.9;
```
neb42, line 96: Not sure I understand what the `tenPercentMargin` is for. I'd expect this method to give `expiresIn` in milliseconds. Returning 90% of the value isn't clear.

neb42:
I think something like this would clean the refresh `useEffect` up. I don't think you really need to be checking the expiration time. We can just blindly refresh it.

```
  React.useEffect(() => {
    const refreshToken = async () => {
      if (!discovery) return;
      if (!tokenResponse) return;
      if (!tokenResponse.expiresIn || tokenResponse.expiresIn <= 0) return;

      const refreshConfig = {
        clientId,
        scopes,
        extraParams: {},
      };

      try {
        const resp = await tokenResponse.refreshAsync(refreshConfig, discovery);
        setTokenResponse(resp);
      } catch {
        setTokenResponse(undefined);
      }
    };

    let interval: number | undefined;
    if (discovery && tokenResponse?.expiresIn) {
      interval = setInterval(refreshToken, tokenResponse.expiresIn / 3);
    }

    return () => {
      if (interval) clearInterval(interval);
    };
  }, [JSON.stringify(discovery), tokenResponse?.expiresIn]);
```

frontend/apps/core-app/src/components/AuthWrapper/useTokenResponse.ts
```diff
@@ -69,29 +69,44 @@ export const useTokenResponse = (): [
...
+
+    if (refreshTokenInterval.current)
+      window.clearInterval(refreshTokenInterval.current);
+
+    if (
+      tokenResponse &&
+      tokenResponse.expiresIn &&
+      tokenResponse.expiresIn > 0
+    ) {
+      refreshTokenInterval.current = window.setInterval(refreshToken, 1000);
```
internetti, line 102: so

PR #354, opened 2022-01-03
COR-437: redirect to form after saving it

internetti:


frontend/apps/pwa/src/features/former/FormerContainer.tsx
```diff
@@ -156,7 +165,10 @@ export const FormerContainer: FC = (props) => {
     if (ownerForm) {
       dispatch(formerActions.saveForm());
     } else if (formDefinition) {
-      dispatch(postForm(formDefinition));
+      const postFormPromise = dispatch(postForm(formDefinition)).unwrap();
+      postFormPromise.then((data) => {
+        history.push(`/browse/forms/${data.id}`);
+      });
```
neb42, line 171: I think the async/await syntax is more readable and more consistent with modern day standards.

PR #355, opened 2022-01-04
COR-436: redirect to record after submitting, refactoring

internetti:


frontend/apps/pwa/src/features/recorder/RecordEditorContainer.tsx
```diff
@@ -0,0 +1,179 @@
...
+          recorderActions.selectRecord({
+            recordId: currentRecord.ownerId,
+          }),
+        );
+      }
+    } else {
+      const recordResponse = await dispatch(postRecord(recordsToPost)).unwrap();
+      dispatch(recordActions.addMany(recordResponse));
+      history.push(`/browse/records/${recordResponse[0].id}`);
+    }
```
internetti, line 150: this is the most relevant change

frontend/apps/pwa/src/reducers/recorder.ts
```diff
@@ -1,54 +1,68 @@
...
-import { recordGlobalSelectors } from '../../reducers/records';
-import client from '../../app/client';
+import { RootState } from '../app/store';
+import client from '../app/client';
 
-export interface FormValue {
+import { FormInterface, selectFormOrSubFormById, selectRootForm } from './form';
+import {recordGlobalSelectors} from "./records";
+
+export interface FormValue extends Omit<Record, 'databaseId'> {
```
internetti, line 16: this type felt redundant, opinions welcome ^^

neb42:
When refactoring components to use the pattern in the style guide we should keep their files together in the directory.
In this pr you added `RecordEditorContainer` in `features/recorder`. I think this should instead be:

```
features/
  recorder/
    RecordEditor/
      index.ts
      RecordEditor.container.tsx
      RecordEditor.component.tsx
```

This way whoever uses `RecordEditor` can be thought of as a the public component with the container/component files being private. By having both the component and container in `features/recorder` it suggests that they are both used elsewhere in the application.

frontend/apps/pwa/src/features/browser/RecordField.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { Link } from 'react-router-dom';
+import format from 'date-fns/format';
+
+type RecordFieldProps = {
+  field: FieldDefinition;
+  value: any;
+  subRecords: Record[] | undefined;
+};
+export const RecordField: FC<RecordFieldProps> = (props) => {
+  const { field, value, subRecords } = props;
```
neb42, line 12: I assume this came from copy/pasting, but we should destruct props like `= ({ field, value, subRecords }) => {

frontend/apps/pwa/src/features/browser/RecordField.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  function renderField(f: FieldDefinition, v: any) {
```
neb42, line 14: I think using arrow functions are typically better as they have block scoping, clearer currying syntax, and align with modern js practices better.

frontend/apps/pwa/src/reducers/recorder.ts
neb42: I find this a little difficult to interpret as we're both extending and omitting. I think this is worth keeping if there is a strong relationship between Record and FormValue. ie. if we add a field to Record we will typically want it in FormValue.

frontend/apps/pwa/src/features/browser/RecordField.tsx
internetti: completely agree, copy paste issues ...

frontend/apps/pwa/src/components/FieldEditor/FieldEditor.tsx
```diff
@@ -0,0 +1,52 @@
...
+  WeekFieldEditor,
+  SingleSelectFieldEditor,
+  SubFormFieldEditor,
+  ReferenceFieldEditor,
+} from '../formFields';
+
+export const FieldEditor: FC<FieldEditorProps> = (props) => {
+  const {
+    field: { fieldType },
+  } = props;
```
neb42, line 20:
```
export const FieldEditor: FC<FieldEditorProps> = ({ field: { fieldType }) => {
```

neb42: ah sorry you need the full props below

frontend/apps/pwa/src/components/formFields/MultiSelectFieldEditor.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <select
+        className="form-control bg-dark text-light border-secondary"
+        id={field.id}
+        value={value || []}
+        multiple
+        onChange={(event) => {
+          const { options } = event.target;
+          const selected = Object.entries(options).filter((o) => o[1].selected);
+          setValue(selected.map((s) => s[1].value));
+        }}
```
neb42, line 27: This should live in an `handleOnChange` function

frontend/apps/pwa/src/components/formFields/types.ts
```diff
@@ -0,0 +1,12 @@
+import { FieldDefinition } from 'core-api-client';
+
+import { FormValue } from '../../reducers/recorder';
+
+export type FieldEditorProps = {
+  field: FieldDefinition;
+  value: string | string[] | null;
+  setValue: (value: string | string[] | null) => void;
```
neb42, line 8: I think a more common naming convention here is `onValueChange`. The component doesn't know that you're "setting" a value, it just knows that you're running something when the value changes.

neb42:
In regard to `frontend/apps/pwa/src/components/formFields/helpers.tsx` I think helper files are generally pretty bad as they quickly grow out of control. In this case some functions are common across fields and some are field specific. As a quick fix I would make a directory, `frontend/apps/pwa/src/components/formFields/helpers`, and add each helper function as a file.

frontend/apps/pwa/src/features/recorder/RecordEditor.component.tsx
```diff
@@ -0,0 +1,77 @@
+import React, { FC } from 'react';
+import { FieldDefinition, FieldValue } from 'core-api-client';
+
+import { FormValue } from '../../reducers/recorder';
+import { FieldEditor } from '../../components/FieldEditor/FieldEditor';
+
+export type RecordEditorProps = {
```
neb42, line 7: Does this need to be exported? If not I'd just call it `Props`

frontend/apps/pwa/src/features/recorder/RecordEditor.component.tsx
```diff
@@ -0,0 +1,77 @@
...
+  fields: FieldDefinition[];
+  values: FieldValue[];
+  setValue: (key: string, value: any) => void;
+  selectSubRecord: (subRecordId: string) => void;
+  addSubRecord: (ownerFieldId: string) => void;
+  subRecords: { [key: string]: FormValue[] };
+  saveRecord: () => void;
```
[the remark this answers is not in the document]
neb42: And when using the component typically `handle<Action>` although this does differ on the use case.

eg. `<RecordEditorComponent onSaveRecord={handleSaveRecord} />`

frontend/apps/pwa/src/features/recorder/index.ts
```diff
@@ -0,0 +1 @@
+export { RecordEditorContainer } from './RecordEditor.container';
```
neb42, line 1:
```
export { RecordEditorContainer as RecordEditor  } from './RecordEditor.container';
```
Components using RecordEditor don't care that it's a container. Exporting like this allows us to refactor it to not use a container in the future if we want.

frontend/apps/pwa/src/features/recorder/RecordEditor.component.tsx
```diff
@@ -0,0 +1,77 @@
...
+              </div>
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
+  );
+};
+
+export default RecordEditorComponent;
```
neb42, line 77: I don't think this line is needed as we're doing a named export

internetti: ah forgot to delete it

PR #356, opened 2022-01-04
DS-42: typography setup

internetti:


frontend/apps/core-app/src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
internetti, line 1: was added automatically

frontend/apps/core-app/tsconfig.json
```diff
@@ -2,10 +2,14 @@
   "extends": "expo/tsconfig.base",
   "compilerOptions": {
     "strict": true,
-    "paths": {
-      "core-api-client": ["../../packages/core-api-client/src"],
-      "core-auth": ["../../packages/core-auth/src"],
-      "core-design-system": ["../../packages/design-system/src"],
-    }
-  }
+    "allowSyntheticDefaultImports": true,
```
internetti, line 5: this was changed when I ran yarn test ...

neb42:
I think it's worth adding the font loading to the app as well

frontend/apps/core-app/tsconfig.json
neb42: hmm strange, I'll try running it myself and seeing what happens. I know `react-scripts` changes `tsconfig.json` but I wouldn't expect expo or jest to.

frontend/packages/design-system/src/themes/Text.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export default {
+  defaultProps: {
+    letterSpacing: '0em',
+    textAlign: 'auto',
+  },
+  variants: {
+    display: {
+      fontFamily: fontConfig.Roboto['700'].normal,
+      fontSize: '44px',
```
neb42, line 11: In `fontConfig` I would typically have a `fontSize` key with the font sizes. IBM use rem units for their's which I quite like. Even though we can't use rem for react native we could set a base font size `1` (typically 16px) and define each font size from that, eg:
```
fontSizes: {
 '0.875': '14px',
  '1': '16px',
  '1.125': '18px',
}
```
we could event compute this:
```
const baseFontSize = 16;
const scales = [0.875, 1, 1.125];
const fontSizes = scales.reduce((acc, cur) => ({ ...acc, [cur]: `${cur * baseFontSize}px` }), {});
```

Perhaps this is overkill and while the typography is still being worked on we can leave it hard coded and discuss this with the designers.
https://www.carbondesignsystem.com/guidelines/typography/overview/

frontend/packages/design-system/src/tokens/fontConfig.ts
```diff
@@ -0,0 +1,11 @@
+export default {
+  Roboto: {
```
neb42, line 2: I'd nest this under `fontFamily` in case we want to add any other tokens around typography (eg. font sizes)

neb42: We may also have multiple font families

internetti: > We may also have multiple font families
they could live on the same level as Roboto, couldn't they?

neb42: Yeah other font families would, but not font sizes/letter spacings/weights/line heights. Not saying we need all of this, but it helps for it to be extensible.

frontend/packages/design-system/src/themes/Text.ts
internetti: I'm not sure how useful this is if the base font size is not configurable. also not sure about the 0.875 keys, native base only seems to work with xs, sm and such

frontend/packages/design-system/src/tokens/typography.ts
```diff
@@ -0,0 +1,55 @@
...
+    heading: 'Roboto',
+    title: 'Roboto',
+    bodyText: 'Roboto',
+    inline: 'Roboto',
+    date: 'Roboto',
+    label: 'Roboto',
+    caption: 'Roboto',
+  },
+  letterSpacings: {
+    xs: '0em',
```
neb42, line 49: Can we use `em` with react native? Does letter spacing get used anywhere?

internetti: no, I'll switch to px, I just copied from figma.
I'd say it's used, since the default values of native base are different from what the designers specified

neb42:
I'm not sure why `frontend/apps/core-app/src/react-app-env.d.ts` and the `tsconfig.json` changes has been made. They seem to have been caused by `react-scripts`, but the app doesn't use `react-scripts`. Can you revert the changes and try again?

internetti:
> I'm not sure why `frontend/apps/core-app/src/react-app-env.d.ts` and the `tsconfig.json` changes has been made. They seem to have been caused by `react-scripts`, but the app doesn't use `react-scripts`. Can you revert the changes and try again?

tried again after reverting these changes, seems to have worked, no idea what I did there

PR #357, opened 2022-01-06
COR-388: access token in session store

internetti:


frontend/packages/core-auth/src/components/AuthWrapper.tsx
```diff
@@ -45,6 +48,13 @@ const AuthWrapper: React.FC<AuthWrapperProps> = ({
     browser,
   );
 
+  // store tokenresponse in session storage
+  useEffect(() => {
+    if (tokenResponse?.accessToken) {
+      sessionStorage.setItem(injectToken, JSON.stringify(tokenResponse));
```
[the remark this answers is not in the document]
internetti: but it did work when I tried it?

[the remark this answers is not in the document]
neb42: hmm yeah didn't realise `stringify` will take the properties to an object. But `getSessionStorage(injectToken)` can't be creating a `TokenResponse`. It's just parsing an object.

frontend/packages/core-auth/src/components/AuthWrapper.tsx
```diff
@@ -45,6 +48,13 @@ const AuthWrapper: React.FC<AuthWrapperProps> = ({
...
+    }
+  }, [tokenResponse?.accessToken]);
```
neb42, line 56: This isn't always `accessToken`. It depends on `injectToken`. I would take this block outside of it's useEffect and whenever we have `accessToken` in useEffect deps,  we just use `token`.
```
const token = (() => {
      switch (injectToken) {
        case 'access_token':
          return tokenResponse?.accessToken ?? '';
        case 'id_token':
          return tokenResponse?.idToken ?? '';
        default:
          return '';
      }
    })();
  ```

File: frontend/packages/core-auth/src/components/AuthWrapper.tsx [hunk 3]
```diff
@@ -78,30 +92,36 @@ const AuthWrapper: React.FC<AuthWrapperProps> = ({
...
+      case 'access_token':
+        setToken(tokenResponse?.accessToken ?? '');
+        break;
+      case 'id_token':
+        setToken(tokenResponse?.idToken ?? '');
+        break;
+      default:
+        setToken('');
+    }
   }, [tokenResponse?.accessToken, tokenResponse?.idToken]);
```
neb42, line 105: I think this is unnecessary. I think calculating the value of `token` on each render is very inexpensive and adding it into a state variable adds too much complexity.

frontend/packages/core-auth/src/components/AuthWrapper.tsx
```diff
@@ -45,16 +60,40 @@ const AuthWrapper: React.FC<AuthWrapperProps> = ({
     browser,
   );
 
+  // Run onTokenChange callback
+  // store token in session storage or remove it when undefined
+  const applyToken = (tr: TokenResponse | undefined) => {
```
neb42, line 65: Rather than having this as a function we could have it in a useEffect, like:

```
React.useEffect(() => {
 ....
}, [JSON.stringify(tokenResponse)]);

File: frontend/packages/core-auth/src/components/AuthWrapper.tsx [hunk 5]
```diff
@@ -157,7 +183,7 @@ const AuthWrapper: React.FC<AuthWrapperProps> = ({
           customLoginComponent({ login: handleLogin })
         ) : (
           <button type="button" onClick={handleLogin}>
-            Login
+            Login {tokenResponse?.accessToken}
```
neb42, line 186: We probably don't want this

frontend/packages/core-auth/src/components/AuthWrapper.tsx
```diff
@@ -8,8 +8,21 @@ import {
   AuthWrapperProps,
   CodeChallengeMethod,
   ResponseType,
+  TokenResponseConfig,
 } from '../types/types';
 import { TokenResponse } from '../types/response';
+import { getSessionStorage } from '../utils/getSessionStorage';
+
+const createTokenResponse = (trc: TokenResponseConfig) => {
```
neb42, line 16: I'd probably move this to a static method on `TokenResponse`. Like `fromJSON`. It also doesn't need to use a `let`.

```
  if (!trc) return undefined;
  try {
    return new TokenResponse(trc);
  } catch (e) {
    return undefined;
  }
```

frontend/packages/core-auth/src/components/AuthWrapper.tsx
```diff
@@ -142,8 +156,20 @@ const AuthWrapper: React.FC<AuthWrapperProps> = ({
...
+  // Update logged in status accordingly
+  useEffect(() => {
+    if (tokenResponse) {
+      if (!isLoggedIn) {
+        setIsLoggedIn(true);
+      }
+    } else if (isLoggedIn) {
+      setIsLoggedIn(false);
+    }
+  }, [tokenResponse, isLoggedIn]);
```
neb42, line 170: `JSON.stringify(tokenResponse)`

frontend/packages/core-auth/src/components/AuthWrapper.tsx
```diff
@@ -142,8 +156,20 @@ const AuthWrapper: React.FC<AuthWrapperProps> = ({
       if (refreshTokenInterval.current)
         clearInterval(refreshTokenInterval.current);
     };
-  }, [tokenResponse?.refreshToken, tokenResponse?.expiresIn]);
+  }, [tokenResponse?.refreshToken, tokenResponse?.expiresIn, discovery]);
```
neb42, line 159: `JSON.stringify(discovery)`

frontend/packages/core-auth/src/utils/getSessionStorage.ts
```diff
@@ -0,0 +1,6 @@
+export const getSessionStorage = (key: string): any | undefined => {
```
neb42, line 1: This name should indicate that it's getting JSON. Maybe, `getJSONFromSessionStorage`?

frontend/packages/core-auth/src/components/AuthWrapper.tsx
internetti: that's where it was originally. but for a reason that I did not find, that use effect was not always triggered when I expected it to because the auth wrapper was not rerendered every time the tokenResponse got updated by the refresh function.
scenarios in which that happened: 
* reload page, wait for token to refresh, next request fails
* trigger a request, wait for token to refresh, then reload

frontend/packages/core-auth/src/components/AuthWrapper.tsx
internetti: oh yeah sorry, that was just for testing something

frontend/packages/core-auth/src/components/AuthWrapper.tsx
internetti: I'll see if the stringify solves that problem

frontend/packages/core-auth/src/components/AuthWrapper.tsx
internetti: what's the pattern behind this? anytime it's an object? even when no single properties are being addressed?

internetti: @neb42

neb42: When useEffect checks is dependencies have changed it does a shallow comparison. So doing a shallow comparison of an instance of `TokenResponse` will always be different on each render. Because of this you want to always compare something that will do a shallow comparison properly (eg. string, number, boolean, null, undefined).

When deciding between `tokenResponse?.refreshToken` or `JSON.stringify(tokenResponse)` it's really about reasoning why you want the useEffect block to run. When you write the token response to local storage you want to write the entire object so you want to run it if any of `JSON.stringify(tokenResponse)` changes. In this block we don't want to reset the refresh interval if the access token changes, because it doesn't use the access token. It's using the refresh token, so `tokenResponse?.refreshToken` makes sense.

For `discovery` here we're using the full discovery object so we want to stringify that.

internetti: But discovery doesn't change right? it's either undefined or in it's final state, so we're testing for pure existance in that case. wouldn't it then be okay to not use JSON.stringify?

neb42: If we want to test for the existence of `discovery`, we should do something like `Boolean(discovery)`. Using just `discovery`, I'd expect this to run on every rerender because it would check if `discoveryNew === discoveryOld` which should return false every time.

I'd say we do want to check the full discovery object because of this line:
```
 const resp = await tokenResponse.refreshAsync(refreshConfig, discovery);
```
If discovery changes (even though it shouldn't) we want to rerun this useEffect so that this line uses the new discovery.

frontend/packages/core-auth/src/components/AuthWrapper.tsx
```diff
@@ -142,8 +123,46 @@ const AuthWrapper: React.FC<AuthWrapperProps> = ({
...
+          case 'id_token':
+            return tokenResponse?.idToken ?? '';
+          default:
+            return '';
+        }
+      })();
+      onTokenChange(token);
+      sessionStorage.setItem(injectToken, JSON.stringify(tokenResponse));
+    }
+  }, [tokenResponse?.accessToken]);
```
neb42, line 152: I think the deps here should be `JSON.stringify(tokenResponse)` because of the line `sessionStorage.setItem(injectToken, JSON.stringify(tokenResponse))` and if `injectToken` was `id_token`

frontend/packages/core-auth/src/utils/__tests__/getSessionStorage.test.js
```diff
@@ -0,0 +1,33 @@
+import { getJSONFromSessionStorage } from '../getJSONFromSessionStorage';
+
+describe('utils/getSessionStorage', () => {
```
neb42, line 3: Maybe add a test for when it tries to get a key that isn't valid json

PR #358, opened 2022-01-11
Eas build

neb42:
This sets up `frontend/apps/core-app` to use [EAS build](https://docs.expo.dev/build/introduction/) (and eventually the submit functionality). 

* Update `core-app` and `storybook` to expo 44 and ensure all expo packages are at the correct version (running `expo upgrade` will do this)
* Remove old `native` app from `yarn workspaces`. It was causing issues with the build. Seen as it currently exists in order to preserve the code I thought it best to just remove it from workspaces.
* Various bits of configuration


frontend/apps/core-app/eas.json
```diff
@@ -0,0 +1,39 @@
+{
+  "cli": {
+    "version": ">= 0.43.0"
+  },
+  "build": {
+    "development": {
+      "distribution": "internal",
+      "releaseChannel": "development",
+      "env": {
+        "REACT_APP_OIDC_ISSUER": "",
```
internetti, line 10: aren't those currently generated to contain the local tunnels?

neb42: When we run this locally it will use the `.env` file. This is for when we build an `apk` and want it pointing at a deployed service. These should be populated, but I don't have any values for them yet.

frontend/apps/core-app/package.json
```diff
@@ -54,5 +65,25 @@
     "collectCoverageFrom": [
       "src/**/*.{js,jsx,ts,tsx}"
     ]
+  },
+  "expo-yarn-workspaces": {
+    "symlinks": [
+      "@unimodules/core",
```
internetti, line 71: what are the criteria for a package to be listed here?

neb42: Good catch, I copied this from an example. I need to dig into this a bit more

neb42: The docs on this are [here](https://github.com/expo/expo/tree/master/packages/expo-yarn-workspaces#symlinks). What I'd say is if a build fails and it complains about not being able to find a package or an issue with a package, then we should try adding it in here. It's not clear what should be here without digging into it's source code.

I have cleaned up the list though.

internetti (commented):
how exactly can this be tested?
the scripts don't contain one for core-app, the package.json in frontend doesn't have a script to build the core-app, running yarn start in the core-app folder runs the app, but it's broken

frontend/apps/core-app/package.json
internetti: yeah, I had seen that, didn't find it very explanatory though. 
what made you choose the ones to keep/remove?

neb42: Kept the ones that are in our package.json (assuming they were in the example snippet for a reason), then did trial and error on the others.

I think what the docs mean is that some packages will look for a file in the path` ./node_modules/<package>/file.ext`. With yarn workspaces these files live at `../../../node_modules/<package>/file.ext`. If it searches for the file where it expects it then it will get an error. Adding the package to the symlink section sets up a symlink from one node_modules/<package> to the other. I think this issue is specific to react native because some of files will be android/ios files.

neb42:
@internetti 
> how exactly can this be tested? the scripts don't contain one for core-app, the package.json in frontend doesn't have a script to build the core-app, running yarn start in the core-app folder runs the app, but it's broken

I think the way to "test" this is to have a build step in our CI pipeline, so that whenever we push to main we build a development version of the app. I don't think it makes sense to have regular tests for this.

In terms of it being broken, do you have the `.env` in place? It seems to be missing from the init secrets command. I've added it back in. After doing that it worked for me.

neb42:
There's a lot of warnings showing up for `react-native-gesture-handler`. They don't seem to be causing any issues and there is a commit that fixes these. This commit hasn't been released yet, but when it is we can update to remove them.

internetti:
@neb42 if the core-app is the one we want to go further with, could you add a make script for it?

PR #359, opened 2022-01-19
COR-456: Fix multiselect field rendering in record editor

neb42:


neb42:
The branch name here is wrong. The pr title is correct

PR #363, opened 2022-01-21
Cor 442 add yes no field type

neb42:
I've left the record rendering as the string "true" or "false". There should be a better visual indicator of this, but I think this is fine for the purposes of the pwa.

### DIFF
--- a/pkg/api/types/validation/form.go
+++ b/pkg/api/types/validation/form.go
@@ -35,6 +35,8 @@ const (
 	errSelectOptionNameRequired              = "Name of the option is required"
 	errSelectOptionsRequired                 = "At least 1 option must be specified"
 	errSubFormCannotBeKeyOrRequiredField     = "Sub form fields cannot key marked as key or required fields"
+	errBooleanFieldCannotBeKey               = "Boolean fields cannot key marked as key field"
+	errBooleanFieldMustBeRequired            = "Boolean fields must be required"
 	fieldCodeMaxLength                       = 32
 	fieldCodeMinLength                       = 1
 	fieldNameMaxLength                       = 128
@@ -196,6 +198,18 @@ func ValidateFieldDefinition(field *types.FieldDefinition, path *validation.Path
 	// validates that the field name is a valid form name if the field is a subform field
 	if field.FieldType.SubForm != nil {
 		result = append(result, ValidateFormName(field.Name, namePath)...)
+	}
+
+	if field.FieldType.Boolean != nil {
+		// Validates that boolean field values are required
+		if !field.Required {
+			result = append(result, validation.Invalid(requiredPath, field.Required, errBooleanFieldMustBeRequired))
+		}
+
+		// Validates that boolean field values are not key fields
+		if field.Key {
+			result = append(result, validation.Invalid(keyPath, field.Key, errBooleanFieldCannotBeKey))
+		}
 	}
 
 	// validates the field type
--- a/pkg/api/types/validation/form.go
+++ b/pkg/api/types/validation/form.go
@@ -2,13 +2,14 @@ package validation
 
 import (
 	"fmt"
+	"reflect"
+	"regexp"
+	"strings"
+
 	"github.com/nrc-no/core/pkg/api/types"
 	"github.com/nrc-no/core/pkg/utils/sets"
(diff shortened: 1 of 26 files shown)


pkg/api/types/validation/form.go
```diff
@@ -461,3 +465,9 @@ func ValidateSelectOptions(options []*types.SelectOption, path *validation.Path)
 
 	return result
 }
+
+func ValidateFieldTypeBoolean(ftBoolean *types.FieldTypeBoolean, path *validation.Path) validation.ErrorList {
```
ludydoo, line 469: Just a thought, but perhaps that a boolean field should not be a "key" field ?

pkg/sqlmanager/reader.go
```diff
@@ -371,6 +387,23 @@ func getIntValue(value interface{}) (types.StringOrArray, error) {
 	}
 }
 
+func getBooleanValue(value interface{}) (types.StringOrArray, error) {
+	switch t := value.(type) {
+	case bool:
+		return types.NewStringValue(strconv.FormatBool(t)), nil
```
ludydoo, line 393: Make sure that strconv.FormatBool does not return an uppercase "True" ;)

pkg/api/types/validation/form.go
neb42: I also think it should be required as there's no clear UI difference between false and null. In the frontend I've made it always default to false, but this isn't enforced on the backend. Do you think I should add that constraint here too?

ludydoo: I would think so.. I can't imagine a usecase where a boolean would be part of the key.. If we get some demands for it, we could "loosen" the constraint later on?

internetti (changes requested):
I cannot save a form with a checkbox in it, there's no corresponding network request
the save callback from the formerContainer gets undefined as formdefinition and the boolean field is not stored in the state with the rest of the form

pkg/sqlmanager/reader.go
neb42: I think the test checks this

neb42:
> I cannot save a form with a checkbox in it, there's no corresponding network request the save callback from the formerContainer gets undefined as formdefinition and the boolean field is not stored in the state with the rest of the form

Weird, I'd already fixed this but it hadn't been committed. I've pushed up a fix now.

internetti (commented):
if not clicking the checkbox is automatically translated to false, why do we need to mark the field as required?

ludydoo:
> if not clicking the checkbox is automatically translated to false, why do we need to mark the field as required?

Good point.. There is also 3 statuses for checkboxes. "Checked" "Not Checked" and "Indeterminate". The "indeterminate" one is the one with the horizontal bar in it.

https://css-tricks.com/indeterminate-checkboxes/

neb42:
> if not clicking the checkbox is automatically translated to false, why do we need to mark the field as required?

If we didn't force it to be required this could lead to inconsistent states between the frontend and the api. If I hit the api directly I could create a checkbox with a null value, but the frontend doesn't know how to interpret it and it doesn't make much sense conceptually. By making it required we ensure that it can only ever be true or false.

ludydoo:
Yeah it's quite ambiguous "false" or "null", but the html checkbox has support for displaying nulls differently than false. Though, very little users will know about that ! Manses sense to force required for now

neb42:
> Good point.. There is also 3 statuses for checkboxes. "Checked" "Not Checked" and "Indeterminate". The "indeterminate" one is the one with the horizontal bar in it.
> 
> https://css-tricks.com/indeterminate-checkboxes/
> 
> Probably that we "could" (should?) allow "nulls"/"undefined" for non-required yes/no fields by allowing the "indeterminate" status? I'm thinking that in the future, when we start implementing "relevance rules" for fields, we'll need to have non-required checkboxes!

I think the use case of intermediate checkboxes isn't clear. I think they become useful if we were to implement checkbox groups, with children (think of selecting emails in gmail). So I think they fall out of the scope of this pr.

If we were to implement them I probably wouldn't use `null` as that is not what `null` means. I'd probably change the value to `0,1,2` or an enum of some sort.

ludydoo:
Looks good to me, @internetti have you resolved the issues on your side ?

internetti (approved):

PR #366, opened 2022-01-26
COR-450 add validation for recipient forms

ludydoo:
Added validation for "Recipient" forms

The validation happens in two parts
1. The static validation, things we can infer without having to make any database calls. That lies in the default "validation" package, where all other validation currently lies.
2. The dynamic validation, that happens for now in the POST /formdefinitions handler. Here, we must make a DB call to check if the "parent" of a recipient form is also a recipient form.

This PR is based off COR-499 branch, so I set the PR target to the COR-499 branch until the other one gets merged into main

### DIFF
--- a/pkg/api/types/validation/recipient_test.go
+++ b/pkg/api/types/validation/recipient_test.go
@@ -3,10 +3,48 @@ package validation
 import (
 	"testing"
 
+	uuid "github.com/satori/go.uuid"
+
 	"github.com/nrc-no/core/pkg/api/types"
 	"github.com/nrc-no/core/pkg/validation"
 	"github.com/stretchr/testify/assert"
 )
+
+func Test_ValidateRecipientForm_SkippedWhenNotRecipientForm(t *testing.T) {
+
+	validFormButInvalidRecipientForm := &types.FormDefinition{
+		Type:       types.DefaultFormType,
+		Name:       "test",
+		DatabaseID: uuid.NewV4().String(),
+		Fields: []*types.FieldDefinition{
+			{
+				Key:      true,
+				Name:     "bla",
+				Required: true,
+				FieldType: types.FieldType{
+					Text: &types.FieldTypeText{},
+				},
+			},
+			{
+				Key:      true,
+				Required: true,
+				Name:     "bli",
+				FieldType: types.FieldType{
+					Text: &types.FieldTypeText{},
+				},
+			},
+		},
+	}
+
+	// assert that the form is valid
+	assert.Equal(t, validation.ErrorList(nil), ValidateForm(validFormButInvalidRecipientForm))
+
+	// change the type to "Recipient"
+	validFormButInvalidRecipientForm.Type = types.RecipientFormType
+
+	// assert that the form is now invalid
+	assert.False(t, ValidateForm(validFormButInvalidRecipientForm).IsEmpty())
+}
 
 func Test_validateRecipientFormHasSingleKeyField(t *testing.T) {
 	tests := []struct {
--- a/pkg/api/types/validation/recipient_test.go
+++ b/pkg/api/types/validation/recipient_test.go
@@ -1,0 +1,84 @@
+package validation
+
+import (
+	"testing"
+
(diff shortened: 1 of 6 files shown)


pkg/api/types/validation/recipient_test.go
```diff
@@ -0,0 +1,84 @@
...
+				FieldType: types.FieldType{
+					Text: &types.FieldTypeText{},
+				},
+			},
+			wantFieldIndex: 0,
+			wantErrs:       nil,
+		},
+		{
+			name: "with multiple key field",
+			fields: []*types.FieldDefinition{
```
internetti, line 50: could we also add a test for testing if that is valid for non-recipients? I don't see where this example form is defined as a recipient form, so this might as well be valid

ludydoo: I`m not sure how to write this test.. 
But this validation is skipped for non-recipient forms. Can you clarify what you would change ?

```
// ValidateRecipientForm validates recipient forms
func ValidateRecipientForm(form *types.FormDefinition) validation.ErrorList {
	var result validation.ErrorList
	if form.Type != types.RecipientFormType {
		return result
	}
         ...
```

internetti: okay, true, that doesn't make sense then. but we could add a test that ValidateRecipientForm really skips the validation/returns an empty error list when the form is not a recipient form?

pkg/server/formsapi/handlers/form/create.go
```diff
@@ -56,6 +62,47 @@ func (h *Handler) Create() http.HandlerFunc {
 	}
 }
 
+func validateRecipientFormParent(
```
internetti, line 65: shouldn't this also live in validation/recipient.go?

internetti: @ludydoo

pkg/api/types/validation/recipient_test.go
ludydoo: Sure, will do !

pkg/server/formsapi/handlers/form/create.go
ludydoo: @internetti the problem with this one is that it requires an access to the database. All other validation happening in validation/recipient.go is "static" (does not need any service call, etc.). Do you think it would be cleaner if it would be in the validation/recipient.go ?

ludydoo: This piece of code should probably live in some kind of controller/validation webhook I think. But we don't have that layer yet !

neb42: What about passing the store to `ValidateForm`?

neb42:
lgtm, but @internetti I'll let you finish before approving

pkg/api/types/validation/recipient_test.go
internetti: @ludydoo it's a detail, but in the test you added, you test ValidateForm instead of ValidateRecipientForm, is there a reason for that?

pkg/server/formsapi/handlers/form/create.go
ludydoo: @neb42 that would work, though it would require all tests to inject the store as well. Imho I really think this should be on a controller layer of some kind.

pkg/api/types/validation/recipient_test.go
ludydoo: @internetti I wanted the test to also validate that the ValidateRecipientForm was called by the main "ValidateForm"

neb42:
@ludydoo I don't think this is handling the case where there are no key fields.

neb42:
I guess this is because of root forms where that is the correct validation. I feel validating a recipient form and a root recipient form are different things, and the root flow shouldn't be exposed to the user.

ludydoo:
@neb42 I thought we agreed to handle this using access control

neb42:
Yeah that rings a bell. Something like:
```
if "super-user" not in userRoles:
    validateRecipientHasKey()
```

PR #367, opened 2022-01-31
COR-451 Recipient seed

ludydoo:
Modified the current seeder to create some recipient forms.
Found out that the seeder is a bit of a mess, the forms are not properly linked together for the Colombia forms.
But this is out of scope of this issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,10 @@ migrate:
 bootstrap: migrate
 	@./scripts/bootstrap.sh
 
+.PHONY: seed
+seed:
+	@./scripts/seed.sh
+
 .PHONY: down
 down:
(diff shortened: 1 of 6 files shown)


neb42:
I'd rename the command and script to `seed_dev`, but otherwise it looks good

ludydoo:
@neb42 moved the `seed` command under the `dev` command (go run dev seed ...), but kept the script filename as is because other files (e.g. `bootstrap.sh`) do not have the `dev_` prefix. We can add a ticket to rename the scripts according to the go commands

neb42:
Ok I'm fine leaving this for now, but I think it's quite important as we'll have some global seeding in production that won't include things like the columbia seeder

PR #371, opened 2022-02-08
Cor 467 forms client refactoring

neb42:
* Refactor client interface to be `client.<resource>.<operation>`
* Split out client code by resource
* Restructure types to be split out my model/client types and then by resource

This does not change the arguments being passed to client methods.


frontend/packages/core-api-client/src/lib/client/Database.ts
```diff
@@ -0,0 +1,31 @@
...
+    return this.restClient.do(
+      request,
+      '/databases',
+      'post',
+      request.object,
+      200,
+    );
+  };
+
+  list = (request: {} | undefined): Promise<DatabaseListResponse> => {
```
internetti, line 28: what do you think about having generic list/create/... functions on the BaseRESTClient? asking because they are mostly the same. could be overkill though?

neb42: I think maybe instead of having generic `list`/`get` etc, we could have methods for `get`/`post` as they roughly translate to the crud methods but are a bit more reusable.

internetti: @neb42 did you mean you would do that in this PR or was that a future plan? i.e. are you still working on this?

neb42: I think in the future, in the interest of getting this merged.

PR #372, opened 2022-02-09
COR-397: form validation client side

internetti:
[COR-397] / [COR-462]

adding client side validation using react-hook-form.
validation for forms as well as records.

missing (will be added in [COR-471] and [COR-472]): 
tests
handling required fields in subforms


frontend/apps/pwa/package.json
```diff
@@ -22,14 +22,16 @@
...
     "@reduxjs/toolkit": "^1.5.1",
     "axios": "^0.23.0",
     "bootstrap": "^5.1.3",
     "bootstrap-icons": "^1.6.1",
     "core-api-client": "*",
     "core-auth": "*",
     "date-fns": "^2.25.0",
-    "lodash": "^4.17.21",
+    "lodash.keyby": "^4.6.0",
+    "lodash.map": "^4.6.0",
```
neb42, line 34: I don't think there's a need for lodash.map. Array map does the same thing

frontend/apps/pwa/src/components/FormPicker.tsx
```diff
@@ -1,34 +1,69 @@
...
 import { useDatabase, useForms } from '../app/hooks';
+import { ValidationForm } from '../reducers/Former/types';
+import { registeredValidation } from '../features/former/validation';
 
 export type FormPickerProps = {
   forms: FormDefinition[];
   formId: string | undefined;
   disabled?: boolean;
   setFormId: (formId: string) => void;
+  register: any;
```
neb42, line 15: Does the library export a type we could use here?

frontend/apps/pwa/src/components/FormPicker.tsx
```diff
@@ -37,16 +72,20 @@ export type FormPickerContainerProps = {
   databaseId: string | undefined;
   formId: string | undefined;
   isRecipientKey: boolean;
-  setForm?: (form: FormDefinition | undefined) => void;
-  setFormId?: (formId: string) => void;
+  // setForm: (form: FormDefinition | undefined) => void;
```
neb42, line 75: Delete this if it isn't used

frontend/apps/pwa/src/components/FormPicker.tsx
internetti: yes, but it lead to a lot of problems :sweat_smile: I'll try using it again

frontend/apps/pwa/src/components/FormPicker.tsx
```diff
@@ -1,34 +1,69 @@
...
+  setFormId,
+  disabled,
+  register,
+  errors,
+}) => {
   const hasForms = forms.length > 0;
+
+  const registerSelectedFieldReference = register(
+    'selectedField.fieldType.reference.formId',
+    registeredValidation.selectedField.fieldType.reference.formId,
```
neb42, line 31: What is this argument doing? It looks like an initial value, but it doesn't seem to match up with what I read in the docs.

frontend/apps/pwa/src/components/FormPicker.tsx
```diff
@@ -55,30 +94,24 @@ const FormPickerContainer: FC<FormPickerContainerProps> = ({
...
-        setFormId(formId);
-      }
-      if (setForm) {
-        const form = forms.find((f) => f.id === formId);
-        setForm(form);
-      }
+    (fID: string) => {
+      setFormId(fID);
+      // const form = forms.find((f) => f.id === fID);
+      // setForm(form);
```
neb42, line 100: Why do we no longer want to se the form?

frontend/apps/pwa/src/components/FormPicker.tsx
```diff
@@ -1,34 +1,69 @@
...
   return (
     <div>
       <select
         disabled={disabled || !hasForms}
-        onChange={(e) => setFormId(e.target.value)}
+        {...registerSelectedFieldReference}
+        onChange={(e) => {
+          setFormId(e.target.value);
+          return registerSelectedFieldReference.onChange(e);
+        }}
```
neb42, line 41: Not suggesting any changes here, but we should maybe think of a better way of structuring this in a container/component pattern. Maybe something like:

Container:
```
const r = register(.....);
const registerProps = {
  ...r,
  onChange: (event: React.ChangeEvent<HTMLSelectElement>) => {
    setFormId(event.target.value);
    return r.onChange(event);
  },
};
 
return <Component registerProps={registerProps} />;
```

Component:
```
return <select {...registerProps}>....</select>;
```

frontend/apps/pwa/src/components/FormPicker.tsx
```diff
@@ -1,34 +1,69 @@
...
         value={formId || ''}
-        className="form-select"
+        className={`form-select ${
+          errors?.selectedField?.fieldType?.reference?.formId && !disabled
```
neb42, line 44: Similarly with the errors, we should extract the specific error in the container. Again, not suggesting changes for now, just for future use.

frontend/apps/pwa/src/components/FormPicker.tsx
```diff
@@ -55,30 +94,24 @@ const FormPickerContainer: FC<FormPickerContainerProps> = ({
     : allForms;
 
   const callback = useCallback(
```
neb42, line 96: Not part of this pr, but I've seen this a lot. `useCallback` is useful if the function is expensive to create or will cause lots of unnecessary rerenders. In most cases `useCallback` has more overhead than just creating the function on each render.

neb42: Same goes for `useMemo` and `React.memo`

neb42:
In order to avoid too much prop drilling we could use `useFormContext` to get `register` and other functions when needed rather than passing them down and polluting the component apis. Maybe not something for this pr, but to consider for when we build out the native app.

frontend/apps/pwa/src/components/formFields/CheckboxFieldEditor.tsx
```diff
@@ -31,7 +30,6 @@ export const CheckboxFieldEditor: React.FC<FieldEditorProps> = ({
           {field.name}
         </label>
       </div>
-      <FieldDescription text={field.description} />
```
neb42, line 34: Why was this removed?

frontend/apps/pwa/src/components/formFields/CheckboxFieldEditor.tsx
```diff
@@ -1,7 +1,6 @@
 import * as React from 'react';
```
neb42, line 1: Does this field not need to be registered?

frontend/apps/pwa/src/components/FormPicker.tsx
internetti: this is passing the validation rules for this field to the register function

neb42: ah yes, I didn't get down to the `registeredValidation` file yet

frontend/apps/pwa/src/components/formFields/CheckboxFieldEditor.tsx
internetti: had to do that to reduce the amount of duplicated code for sonar

frontend/apps/pwa/src/components/formFields/MultilineTextFieldEditor.tsx
```diff
@@ -1,24 +1,30 @@
...
+        errors?.values && errors?.values[field.id] ? 'is-invalid' : ''
+      }`}
+      id={field.id}
+      value={value || ''}
+      {...registerObject}
+      onChange={(event) => {
+        onChange(event.target.value);
+        return registerObject.onChange(event);
+      }}
+      aria-describedby="errorMessages"
```
neb42, line 27: Should this not be the id of the description field? I think it can have multiple entries so we could include both.

frontend/apps/pwa/src/components/formFields/CheckboxFieldEditor.tsx
neb42: I guess it should still be registered, but without validation. I haven't got to it yet, but I assume we're not using react hook forms for the submission? If we were to do that in the future it'd need to be registered.

frontend/apps/pwa/src/components/formFields/WeekFieldEditor.tsx
```diff
@@ -1,38 +1,40 @@
...
-  };
-
+  const expectedLength = 8;
+  const registerObject = register(`values.${field.id}`, {
+    required: { value: field.required, message: 'This field is required' },
+    pattern: {
+      value: /^(?:19|20|21)\d{2}-W[0-5]\d$/,
+      message: 'wrong pattern',
+    },
+    maxLength: { value: expectedLength, message: 'Value is too long' },
```
neb42, line 19: Do we need this if we have the pattern match? I think the message is misleading as the pattern is the priority.

frontend/apps/pwa/src/features/databases/DatabaseEditor.tsx
```diff
@@ -51,7 +56,7 @@ export const DatabaseEditor: FC = (props) => {
                   aria-describedby="nameFeedback"
                 />
                 <div className="invalid-feedback is-invalid" id="nameFeedback">
-                  {_.map(errors, (e) => {
+                  {map(errors, (e) => {
```
neb42, line 59: `errors.map((e) => {`

frontend/apps/pwa/src/features/former/FormerContainer.tsx
```diff
@@ -165,23 +178,53 @@ export const FormerContainer: FC = () => {
   );
 
   const saveField = useCallback(
-    (fieldId: string) => {
-      dispatch(actions.resetFormErrors());
-      dispatch(actions.selectField({ fieldId: undefined }));
+    async (field: FormField) => {
+      const valid = await trigger('selectedField');
+      const errors = customValidation.selectedField(field);
+      // console.log('field errors', errors);
```
neb42, line 184: Delete

frontend/apps/pwa/src/features/former/FormerContainer.tsx
```diff
@@ -165,23 +178,53 @@ export const FormerContainer: FC = () => {
...
+          setError(error.field, {
+            message: error.message,
+          });
+        });
+        return;
+      }
+    }
+
     if (ownerForm) {
+      const valid = await trigger();
```
neb42, line 215: Why are we only running this on subforms?

frontend/apps/pwa/src/features/databases/DatabaseEditor.tsx
internetti: errors is not an array.

frontend/apps/pwa/src/features/former/validation.ts
```diff
@@ -0,0 +1,135 @@
...
+    },
+  },
+};
+
+type CustomError = {
+  field: FieldPath<ValidationForm>;
+  message: string;
+};
+
+export const customValidation = {
```
neb42, line 87: Could this not be part of react form hooks validation? I haven't given any thought to how, but it would be nice if it handled everything.

Maybe we have two instances of react form hooks. One validates the form and one validates the current field.

frontend/apps/pwa/src/components/FormPicker.tsx
internetti: not a good one. will change that.

frontend/apps/pwa/src/features/former/validation.ts
internetti: I didn't think so, because register only works on input and select elements, and we have things that need to be validated that are not bound to one of those, like amount of fields, amount of options, and such.
We could refactor the whole form and validation to work with the control mechanism from react-hook-form, but that would not be a small change.

neb42: Yeah agreed. I think when we build out the native app we should think with react form hooks from the start and try to use it properly. Another one to look at for this custom validation would be io-ts, but I think this suffices for now (I also don't think that would catch things like number of fields).

frontend/apps/pwa/src/features/databases/DatabaseEditor.tsx
neb42: `Object.values(errors).map((e) => {`

frontend/apps/pwa/src/reducers/Former/types.ts
```diff
@@ -1,7 +1,17 @@
...
+  FieldTypeDate,
+  FieldTypeMonth,
+  FieldTypeMultilineText,
+  FieldTypeMultiSelect,
+  FieldTypeQuantity,
+  FieldTypeReference,
+  FieldTypeSingleSelect,
+  FieldTypeText,
+  FieldTypeWeek,
+} from 'core-api-client/src';
```
neb42, line 14: `src` shouldn't be here

frontend/apps/pwa/src/reducers/Former/types.ts
```diff
@@ -15,9 +25,36 @@ export interface FormField {
   subFormId: string | undefined;
   referencedDatabaseId: string | undefined;
   referencedFormId: string | undefined;
-  errors: ErrorMessage | undefined;
 }
 
+export type FieldDefinitionNC = {
```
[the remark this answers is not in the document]
neb42: Maybe `Flat`. Or if sticking with non-circular maybe be explicit with `NonCircular`

frontend/apps/pwa/src/types/errors.ts
```diff
@@ -1,7 +1,7 @@
+import { FieldPath } from 'react-hook-form';
```
neb42, line 1: Is this used?

neb42:
Generally looks good for validation, but there's some work to do to use react form hooks in a cleaner way and to use more of the functionality. We can look at this as we build out the native app.

neb42:
Date, week, and month fields that are required don't fail validation on blank values.

image

internetti:
@neb42 I do get error messages for month, date and week when I don't enter anything. (open form, directly click submit)
did you do anything different? I don't even manage to enter anything that is not a date in these fields, if you did, please explain.

PR #374, opened 2022-02-16
COR-468: input theming

internetti:
[COR-468]

* set up theme for input fields, along with the form controls around it
* add icons for indicating validity of an input
* fix design of buttons (border radius, typography)
* create custom radii, space and size variables to not mess with the native base system
* add new font for labels

Note: 
there is a discussion with the designers about how to handle viewboxes of icons. the result might change a few values.

[COR-468]: https://dteam-nrc.atlassian.net/browse/COR-468?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


frontend/packages/design-system/src/themes/Button.ts
```diff
@@ -80,4 +86,10 @@ export default {
       };
     },
   },
+  sizes: {
+    nrc_10: {
+      height: 'nrc_10',
```
neb42, line 91: Why do we have the `nrc` prefix here? Does this mean when we change `size` we need to do `size="nrc_10"`? If so that doesn't feel right.

frontend/packages/design-system/src/themes/Button.ts
```diff
@@ -1,12 +1,19 @@
+import textTheme from './Text';
+
 export default {
+  defaultProps: {
+    _text: {
+      ...textTheme.variants.button,
```
neb42, line 6: I thought we didn't want to use `defaultProps` for styles?

frontend/packages/design-system/src/themes/FormControlHelperText.ts
```diff
@@ -0,0 +1,13 @@
+import textTheme from './Text';
+
+export default {
+  baseStyle: () => {
+    return {
+      mt: '0.5',
+      _text: {
+        ...textTheme.variants.caption,
+        color: 'neutral.300',
```
neb42, line 9: Why is `color` not part of `textTheme.variants.caption`?

frontend/packages/design-system/src/themes/Button.ts
internetti: yeah that would mean `size="nrc_10"`
I thought it would be better to not overwrite the space and size values of native base and better to add our custom ones, so I needed a new name, and I went with `nrc_` because I thought this design system is supposed to be used nrc-wide.
do you disagree with the custom values, or with the name? or both? :sweat_smile:

frontend/packages/design-system/src/themes/Button.ts
neb42: I forget exactly, but I feel like props are the wrong place for styles as they are easy for the user to override. In other parts of the theme you've used `baseStyle` for styles shared across themes. Maybe that's a better place for this.

frontend/packages/design-system/src/themes/FormControlHelperText.ts
neb42: Yeah fair point. Maybe there should be a default colour (black), but this looks good for these components.

frontend/packages/design-system/src/themes/Button.ts
neb42: I think it's better to override the defaults. Say you're using native base with no theme and you have:
```
<Component size="10" />
```
Then you decide you want to use the NRC theme. After applying the theme your component won't be using the nrc theme without going through the codebase and changing it to `nrc_10`.

I also just think it doesn't read as well and could lead to mistakes where people use `size="10"` as it is more intuitive, but is the incorrect style.

internetti: @neb42 I would agree. but I felt that our spacing/sizing concept was not mature enough to replace it.
native base has a list of 43 different values, a system based on 4. I could replace everything by our system of 5, but implementing the components, I see that we do also have margins and paddings of 2px or 12px and other numbers not divisible by 5. so I'd have to add new values to the theme, but what do I name those? if "1"=5px, "2"=10px, what is 2 pixels? they do work with fractions, but I'm not very happy naming a variable "5/2.5" to arrive at 2px.

if you have an idea on how to integrate these non-5 values into the system name-wise, I'm happy to do that!

neb42: Yeah I guess this is why I don't like using numbers for token values. I think for now leaving the spacing tokens as they are is ok. But I would change the button size. I don't think it should be using numbers. We should stick with `"xs", "sm", "md", "lg"` or a reduced set of that for what we need.

neb42: I think we want to make using our design system as similar to using native base as possible. So if someone reads the native base docs they could more or less use the nrc theme.

frontend/storybook/App.tsx
```diff
@@ -18,9 +20,12 @@ export default function App() {
   const [fontsLoaded] = useFonts({
     Roboto_400Regular,
     Roboto_400Regular_Italic,
+    Roboto_500Medium,
     Roboto_700Bold,
   });
 
+  console.log('THEME', theme);
```
neb42, line 27: Delete

frontend/packages/design-system/src/themes/Input.ts
```diff
@@ -0,0 +1,43 @@
...
+    padding: 3,
+    _android: {
+      ...textTheme.variants.bodyText,
+    },
+    _ios: {
+      ...textTheme.variants.bodyText,
+    },
+    _web: {
+      ...textTheme.variants.bodyText,
+    },
```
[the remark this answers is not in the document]
internetti: @neb42 they might be smaller on a phone?

frontend/packages/design-system/src/themes/Input.ts
```diff
@@ -0,0 +1,37 @@
...
+    backgroundColor: 'white',
+    borderWidth: 1,
+    borderStyle: 'solid',
+    borderColor: 'neutral.300',
+    borderRadius: 'nrc_xs',
+    color: 'neutral.300',
+    padding: 3,
+    _web: {
+      ...textTheme.variants.bodyText,
+    },
```
neb42, line 26: Does this get applied to ios/android if it's under web? Should it not live at the top level?

internetti: I think it would not get applied to ios/android. I'll add it back then?
I would personally also have put it at top level, but this value lives there in the default theming, and if I want to overwrite it, I need to put it here.

neb42: > this value lives there in the default theming
I can't see this in the default theming (https://github.com/GeekyAnts/NativeBase/blob/master/src/theme/components/input.ts)

Does it not work at the top level?

internetti: it didn't when I tried

PR #376, opened 2022-02-17
COR-470: add text area story

internetti:


frontend/storybook/storybook/stories/TextArea/TextArea.stories.js
```diff
@@ -0,0 +1,56 @@
...
+    return (
+      <VStack space={4}>
+        <FormControl>
+          <FormControl.Label>
+            <Text variant="body2">
+              Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
+              eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut
+              enim ad minim veniam, quis nostrud exercitation ullamco laboris
+              nisi ut aliquip ex ea commodo consequat.
+            </Text>
```
neb42, line 19: Why do we have these? It's not really part of the text area

internetti: I considered it to be part of the form controls, as in, the label. I felt it was quite relevant to the look of the text area.
I felt it would be a bit redundant to give variants to the FormControl.Label (since it looks different for input fields), which would be the same as the Text variants, so I used a Text component to demonstrate what it looks like.

neb42: Ah yeah I see. Maybe this is ok for now, but if this style is used more extensively than for text area we should think about adding variants.

frontend/packages/design-system/src/themes/FormControlLabel.ts
```diff
@@ -8,6 +8,23 @@ export default {
...
+      _text: {
+        ...textTheme.variants.body({ level: '2' }),
+      },
+    },
+    sm: {
+      _text: {
+        ...textTheme.variants.label({ level: '1' }),
+      },
+    },
+  },
```
neb42, line 29: I think this is better than using text children

neb42: I'd maybe add a default prop for `size: 'sm'`

PR #378, opened 2022-02-18
COR-466: recipient form validation

internetti:
[COR-466]

validate recipient forms on submit

[COR-466]: https://dteam-nrc.atlassian.net/browse/COR-466?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/apps/pwa/src/features/former/validation.ts
+++ b/frontend/apps/pwa/src/features/former/validation.ts
@@ -1,5 +1,5 @@
 import { FieldPath } from 'react-hook-form';
-import { FieldDefinition, FieldKind } from 'core-api-client';
+import { FieldDefinition, FieldKind, FormType } from 'core-api-client';
 import { RegisterOptions } from 'react-hook-form/dist/types/validator';
 
 import { Form, FormField, ValidationForm } from '../../reducers/Former/types';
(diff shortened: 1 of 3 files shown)


frontend/apps/pwa/src/features/former/validation.ts
```diff
@@ -131,6 +131,24 @@ export const customValidation = {
...
+      if (keyFields.length !== 1) {
+        errors.push({
+          field: 'fields' as FieldPath<ValidationForm>,
+          message: 'Form needs to have exactly 1 key field',
+        });
+      } else if (keyFields[0].fieldType !== FieldKind.Reference) {
+        console.log('key fields', keyFields);
+
+        errors.push({
+          field: 'fields' as FieldPath<ValidationForm>,
```
neb42, line 147: Not important, but I think `as const` should also work here. It'll turn the type from `string` to `fields`.

internetti: seems to work, what benefit would the const version have? just shorter? less imports?

neb42: Yeah, I'm not sure if there's a good reason to use one over the other. I personally use `const`, which as you say uses less imports. But I think both are fine.

PR #379, opened 2022-02-22
Base client rest methods

neb42:
Add get, post, put, delete methods to the base rest client

### DIFF
--- a/frontend/packages/core-api-client/src/lib/client/BaseRESTClient.ts
+++ b/frontend/packages/core-api-client/src/lib/client/BaseRESTClient.ts
@@ -38,6 +38,7 @@ export class BaseRESTClient {
     this.token = token;
   };
 
+  // Deprecated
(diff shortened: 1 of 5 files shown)


frontend/packages/core-api-client/src/lib/client/BaseRESTClient.ts
```diff
@@ -64,4 +65,58 @@ export class BaseRESTClient {
     }
     return clientResponse<TRequest, TBody>(value, request, expectStatusCode);
   }
+
+  public async makeRequest<TRequestBody, TResponseBody>(
```
internetti, line 69: does this actually need to be public?

neb42: No. My thinking was that it may be useful to have access to the generic method, but maybe this can be private for now.

internetti (approved):

PR #380, opened 2022-02-22
COR-485: stubs

internetti:
[COR-485]

[COR-485]: https://dteam-nrc.atlassian.net/browse/COR-485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


frontend/apps/core-app/src/navigators/root.tsx
```diff
@@ -1,35 +1,30 @@
...
-          </View>
-        )}
+        name="RecipientRegistration"
+        component={RecipientRegistrationScreen}
+        options={{
+          title: 'Recipient Registration',
+          headerTitle: 'Recipient Registration',
+        }}
       />
     </Drawer.Navigator>
```
neb42, line 29: I think the drawer should be a single item that renders the recipients stack, defaulting to the list page

neb42: On second thought, this may be useful for development

frontend/apps/core-app/src/screens/RecipientListScreen/RecipientListScreen.component.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <Styles.Container>
+      <Text variant="display">{route.name}</Text>
+      <VStack space={2} width="sm">
+        <Button
+          variant="major"
+          onPress={() =>
+            navigation.navigate('RecipientProfile', {
+              id: '1',
+            })
+          }
```
neb42, line 28: These should be named functions, but probably alright as is seen as they'll go soon

frontend/apps/core-app/src/screens/RecipientListScreen/RecipientListScreen.component.tsx
```diff
@@ -0,0 +1,45 @@
...
+};
+
+export const RecipientListScreenComponent: React.FC<Props> = ({
+  route,
+  navigation,
+}) => {
+  return (
+    <Styles.Container>
+      <Text variant="display">{route.name}</Text>
+      <VStack space={2} width="sm">
```
neb42, line 21: Here the spacing will use the native base spacing, not nrc spacing. To use nrc spacing this would need to be `nrc_x`. We should discuss this again outside of this pr.

frontend/apps/core-app/src/screens/RecipientListScreen/RecipientListScreen.styles.ts
```diff
@@ -0,0 +1,4 @@
+import styled from 'styled-components/native';
+import { Box } from 'native-base';
+
+export const Container = styled(Box)``;
```
neb42, line 4: I know we said to do this in the style guide. Just wondering if it's really necessary.

frontend/apps/core-app/src/screens/RecipientListScreen/__tests__/RecipientListScreen.component.test.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { RootParamList } from '../../../navigators/types';
+
+it('should match the snapshot', () => {
+  const Wrapper: React.FC = () => {
+    const navigation = useNavigation<NavigationProp<RootParamList>>();
+    return (
+      <RecipientListScreenComponent
+        route={{ name: 'RecipientList', params: {}, key: 'key' }}
+        navigation={navigation}
+      />
```
neb42, line 15: To test this I wouldn't normally use the wrapper/hook.

This will be easier to test for any future tests
```
const navigateMock = jest.fn();
const { toJSON } = render(<RecipientListScreenComponent
  route={{ name: 'RecipientList', params: {}, key: 'key' }}
  navigation={{ navigate: navigateMock }}
/>);
```

frontend/apps/core-app/src/navigators/recipients.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { RecipientListScreen } from '../screens/RecipientListScreen';
+import { RecipientProfileScreen } from '../screens/RecipientProfileScreen';
+
+const Stack = createStackNavigator();
+
+export const RecipientNavigator: React.FC = () => {
+  return (
+    <Stack.Navigator initialRouteName="RecipientList">
+      <Stack.Screen
+        name="RecipientList"
```
neb42, line 13: These name strings are used in a lot of places. They should be defined as constants somewhere.

frontend/apps/core-app/App.tsx
```diff
@@ -25,10 +26,31 @@ const App: React.FC = () => {
...
+        RecipientRegistration: 'recipients/register',
+        RecipientProfile: {
+          path: '/:id',
+          parse: {
+            id: String,
+          },
+        },
+      },
+    },
+  };
```
neb42, line 48: Maybe this should live in it's own file. It'll likely grow quite large to be in `App.tsx`

frontend/apps/core-app/src/screens/RecipientListScreen/RecipientListScreen.component.tsx
internetti: generally yeah, but I didn't see the need to pretend like this is even close to a real design ^^

frontend/apps/core-app/src/navigators/recipients.tsx
internetti: oh yeah, thought of doing that, forgot :+1:

frontend/apps/core-app/src/screens/RecipientListScreen/RecipientListScreen.component.tsx
neb42: Yeah, not important for this instance. But I think it is important for defining the spacing in the design system.

frontend/apps/core-app/src/screens/RecipientListScreen/RecipientListScreen.styles.ts
neb42: Whether the entire file is needed

frontend/apps/core-app/src/constants/routes.ts
```diff
@@ -0,0 +1,32 @@
+type Route = {
+  name: string;
+  headerTitle: string;
+  pageTitle: string;
+};
+
+type Routes = {
+  [key: string]: Route;
```
neb42, line 8: Maybe make this more explicit:
```
type Routes = {
  recipientsRoot: Route;
  recipientsList: Route;
  recipientsRegistration: Route;
  recipientsProfile: Route;
};
```

frontend/packages/core-api-client/src/lib/client/Recipient.ts
```diff
@@ -0,0 +1,31 @@
...
+    return this.restClient.post( '/records', recipient);
+  };
+
+  list = (request: RecipientListRequest): Promise<Response<undefined, RecipientList>> => {
+    const { databaseId, formId } = request;
+    const url = `/records?databaseId=${databaseId}&formId=${formId}`;
+    return this.restClient.get(url);
+  };
+
+  get = async (request: RecipientGetRequest): Promise<Response<undefined, Recipient>> => {
```
neb42, line 26: I'd also move these methods away from using a request object.

Either
```
list = (databaseId: string, formId: string)
get = async (databaseId: string, formId: string; recordId: string)
```
or
```
type FormLookup = {
  databaseId: string;
  formId: string;
};

type RecordLookup = {
  databaseId: string;
  formId: string;
  recordId: string;
};
list = (formLookup: FormLookup)
get = async (recordLookup: RecordLookup)
```

neb42: I realise the second option is essentially the same as the request object, but it moves away from the abstract concept of a request to something more concrete. Request can also be confused with http request objects.

frontend/packages/core-api-client/src/lib/types/client/Recipient.ts
```diff
@@ -0,0 +1,8 @@
+import {FieldValue, Record, RecordList} from '../model/Record';
+import {RecordListRequest, RecordGetRequest} from "./Record";
+
+export type RecipientDefinition = FieldValue[]
```
neb42, line 4: I think a `RecipientDefinition` is actually just `Omit<Recipient, 'id'>`. Technically the backend will accept an id, but I think that's bad.

frontend/packages/core-api-client/src/lib/client/Recipient.ts
```diff
@@ -0,0 +1,31 @@
...
+export class RecordClient {
+  restClient: BaseRESTClient;
+
+  constructor(restClient: BaseRESTClient) {
+    this.restClient = restClient;
+  }
+
+  create = (recipient: RecipientDefinition): Promise<Response<RecipientDefinition, Recipient>> => {
+    return this.restClient.post( '/records', recipient);
+  };
```
neb42, line 18: Up for debate, but I prefer the `Response` to be internal to the api and for us to throw an error if it fails. Something like:

```
 create = async (recipient: RecipientDefinition): Promise<Recipient> => {
    const response = await this.restClient.post<RecipientDefinition, Recipient>(
      '/records',
      recipient,
    );
    if (response.success)
      return (response as SuccessResponse<RecipientDefinition, Recipient>).response;
    throw new Error();
  };
```

neb42: Although actually, this method should just forward the request to the record client.

```
constructor(recordClient: RecordClient) {
    this.recordClient = recordClient;
}

return this.recordClient.create(recipient);
```

frontend/apps/core-app/src/screens/RecipientListScreen/RecipientListScreen.styles.ts
internetti: I'm assuming that we'll add more styling to the container later, so I'd keep it for now

frontend/packages/core-api-client/src/lib/client/Recipient.ts
internetti: agreed about keeping responses api-internal. wouldn't throw an error the way things are right now, because the details from the ErrorResponse are needed for validation.
could probably be done by throwing an error too, but as this is blocking multiple tickets and us moving to the native app, I don't think that should be the focus.

frontend/packages/core-api-client/src/lib/client/Recipient.ts
internetti: I'll keep the object format to prevent confusion about which id is which

frontend/packages/core-api-client/src/lib/client/Recipient.ts
internetti: agreed about keeping responses api-internal. wouldn't throw an error the way things are right now, because the details from the ErrorResponse are needed for validation.
could probably be done by throwing an error too, but as this is blocking multiple tickets and us moving to the native app, I don't think that should be the focus.

neb42: Yeah happy to keep it as this for now and refactor later. I would typically include the error messages in the error object.

frontend/packages/core-api-client/src/lib/client/Client.ts
```diff
@@ -5,6 +5,7 @@ import { DatabaseClient } from './Database';
 import { FolderClient } from './Folder';
 import { FormClient } from './Form';
 import { RecordClient } from './Record';
+import {RecipientClient} from "./Recipient";
```
neb42, line 8: Looks like your linting isn't running here

frontend/packages/core-api-client/src/lib/types/model/Record.ts
```diff
@@ -3,6 +3,13 @@ export type FieldValue = {
   value: string | string[] | null;
 };
 
+export type RecordDefinition = {
+  databaseId: string;
+  formId: string;
+  ownerId: string | undefined;
+  values: FieldValue[];
+};
```
neb42, line 11:
```
export type RecordDefinition = Omit<Record, 'id'>;
```

frontend/packages/core-api-client/src/lib/types/client/Recipient.ts
```diff
@@ -0,0 +1,16 @@
...
+} from "./Record";
+
+export type Recipient = Record;
+export type RecipientDefinition = Omit<Recipient, 'id'>
+export type RecipientList = RecordList;
+
+export interface RecipientClientDefinition {
+  create: DataOperation<RecipientDefinition, Recipient>;
+  list: DataOperation<FormLookup , RecipientList>;
+  get: DataOperation<RecordLookup, Recipient>;
```
neb42, line 15: I don't think a `DataOperation` is the correct type here. It works be accident as the methods have a single argument, but would fail with additional arguments.

frontend/packages/core-api-client/src/lib/client/Record.ts
```diff
@@ -16,19 +13,29 @@ export class RecordClient {
...
   }
 
-  create = (request: RecordCreateRequest): Promise<RecordCreateResponse> => {
-    return this.restClient.do(request, '/records', 'post', request.object, 200);
+  create = async (record: RecordDefinition): Promise<Record> => {
+    const apiResponse = await this.restClient.post('/records', record);
+    if (apiResponse.success) {
+      return apiResponse.response as Record
+    }
+    return apiResponse.error
```
neb42, line 21: Returning `apiResponse.error` should give the incorrect type as it's expecting `Promise<Record>`. `apiResponse.error` has the type `any` so typescript allows this.

It should either return `Promise<Record | ErrorThing>` or throw an error.

I would suggest leaving this file as it was and handling the success check in the recipient client.

frontend/packages/core-api-client/src/lib/client/Recipient.ts
```diff
@@ -0,0 +1,27 @@
...
+export class RecipientClient {
+  recordClient: RecordClient;
+
+  constructor(recordClient: RecordClient) {
+    this.recordClient = recordClient;
+  }
+
+  create = (recipient: RecipientDefinition): Promise<Recipient> => {
+    return this.recordClient.create(recipient);
+  };
```
neb42, line 18: Based on my other comment this would become:

```
  create = async (record: RecipientDefinition): Promise<Recipient> => {
    const response = this.recordClient.create({ object: record });
    if (response.success)
      return (response as SuccessResponse<ExampleDefinition, Example>).response;
    throw new APIError((response as ErrorResponse<ExampleDefinition, Example>).error); // TODO define a custom error
  };
```

neb42:
Actually, on my comments about the client: this is a stubs pr. The client shouldn't really be implemented here and should just return some fake data.

frontend/packages/core-api-client/src/lib/types/client/Recipient.ts
internetti: not if we stick to passing arguments as an object

frontend/.eslintignore
```diff
@@ -1,5 +1,5 @@
 node_modules
 dist
 build
-lib
-__generated__
\ No newline at end of file
+core-api-client/lib
```
internetti, line 4: @neb42 maybe another lib folder I forgot?

neb42: That path is wrong, it should be `packages/core-api-client/lib`, but a better path would be `**/lib`

neb42: Same for `dist` and `build` above and probably add an additional `**/node_modules` (keeping the current `node_modules`).

I'm not sure if `**/` will include the current directory

frontend/packages/core-api-client/src/lib/types/client/Record.ts
```diff
@@ -1,12 +1,16 @@
-import { Record, RecordList } from '../model/Record';
+import { Record, RecordList } from '../model';
 
 import { DataOperation, PartialObjectWrapper, Response } from './utils';
 
 export type RecordCreateRequest = PartialObjectWrapper<Record>;
 export type RecordCreateResponse = Response<RecordCreateRequest, Record>;
 
+export type FormLookup = { databaseId: string; formId: string };
```
neb42, line 8: This should be in `types/client/Form.ts`

frontend/packages/core-api-client/src/lib/client/Recipient.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  create = (recipient: RecipientDefinition): Promise<Recipient> => {
+    return Promise.resolve({
+      id: 'id',
+      ...recipient,
+    });
+  };
+
+  list = (args: FormLookup): Promise<RecipientList> => {
+    return Promise.resolve({ items: [] });
```
neb42, line 25: I don't see the use of the `items` key. Could this just be an array?

internetti: it seems to be the current standard of what the api returns. if we extract the items at some point, this could be an array, but if we ever want to do pagination, this might be necessary

frontend/.eslintignore
internetti: **/lib would ignore all lib folders, but the code of the client package is also in a lib folder

internetti: I'll see if I can change that, don't know why it's like that

neb42: > **/lib would ignore all lib folders, but the code of the client package is also in a lib folder

ok then we should just explicitly add each lib folder

frontend/packages/core-api-client/src/lib/client/Recipient.ts
neb42: yeah good shout, this is better for pagination

frontend/packages/core-api-client/src/lib/types/client/Record.ts
internetti: depends on what the structure is here. the name suggests form, yes, but it's going to be used in the record client.

PR #381, opened 2022-02-28
COR-490: accordion component

internetti:
[COR-490]

introducing a simple accordion component into the design system, until native base introduces its own

open questions:
* how much should we invest in styling this?

### DIFF
--- a/frontend/packages/design-system/src/components/Accordion/Accordion.component.tsx
+++ b/frontend/packages/design-system/src/components/Accordion/Accordion.component.tsx
@@ -1,0 +1,24 @@
+import React, { FC, ReactNode } from 'react';
+import { Text, Box, Pressable } from 'native-base';
+
+type Props = {
+  header: string;
+  content: ReactNode;
+};
+
+export const AccordionComponent: FC<Props> = ({ header, content }) => {
+  const [isActive, setIsActive] = React.useState(false);
+
+  return (
+    <Box>
+      <Pressable bg="neutral.200" p="2" onPress={() => setIsActive(!isActive)}>
+        <Text variant="heading">{header}</Text>
+      </Pressable>
+      {isActive && (
+        <Box bg="neutral.100" p="2">
+          {content}
+        </Box>
+      )}
+    </Box>
+  );
+};
--- a/frontend/packages/design-system/src/components/Accordion/index.ts
+++ b/frontend/packages/design-system/src/components/Accordion/index.ts
@@ -1,0 +1,1 @@
+export { AccordionComponent as Accordion } from './Accordion.component';
--- a/frontend/packages/design-system/package.json
+++ b/frontend/packages/design-system/package.json
@@ -11,6 +11,7 @@
   ],
   "dependencies": {
     "native-base": "~3.2.2",
+    "react-native-collapsible": "^1.6.0",
     "react-native-safe-area-context": "~3.3.2",
     "react-native-svg": "~12.1.1",
     "react-native-web": "~0.17.5",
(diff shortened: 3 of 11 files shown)


frontend/packages/design-system/src/components/Accordion/Accordion.component.tsx
```diff
@@ -0,0 +1,24 @@
...
+  header: string;
+  content: ReactNode;
+};
+
+export const AccordionComponent: FC<Props> = ({ header, content }) => {
+  const [isActive, setIsActive] = React.useState(false);
+
+  return (
+    <Box>
+      <Pressable bg="neutral.200" p="2" onPress={() => setIsActive(!isActive)}>
```
neb42, line 14:
```
const handleToggleAccordion = () => setIsActive(!isActive);
....
<Pressable bg="neutral.200" p="2" onPress={handleToggleAccordion}>
```

frontend/packages/design-system/src/components/Accordion/index.ts
```diff
@@ -0,0 +1 @@
+export { AccordionComponent as Accordion } from './Accordion.component';
```
neb42, line 1: Not suggesting to change this, but I feel the `.component` naming pattern is less useful in a design system. Because the components are generally presentational there's less of a need for the container/component design pattern as they are always components.

frontend/packages/design-system/src/components/Accordion/Accordion.component.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { FC, ReactNode } from 'react';
+import { Text, Box, Pressable } from 'native-base';
+
+type Props = {
+  header: string;
+  content: ReactNode;
+};
+
+export const AccordionComponent: FC<Props> = ({ header, content }) => {
+  const [isActive, setIsActive] = React.useState(false);
```
neb42, line 10: `isOpen` is maybe clearer for an accordion

frontend/packages/design-system/src/components/Accordion/Accordion.component.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+  return (
+    <Box>
+      <Pressable bg="neutral.200" p="2" onPress={() => setIsActive(!isActive)}>
+        <Text variant="heading">{header}</Text>
+      </Pressable>
+      {isActive && (
+        <Box bg="neutral.100" p="2">
```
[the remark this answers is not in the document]
internetti: Agree about the color, but if the header has a padding, I'd keep the padding here too, to ensure consistency. 
I wouldn't leave it up to the user of this component to know what the padding of the header is exactly and set the same on their content.
I think more padding is acceptable, but not less, so I'd keep this as a minimum.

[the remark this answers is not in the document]
neb42: Yeah that sounds good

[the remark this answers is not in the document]
internetti: okay, so setting the padding in the DS and bg color in the actual app is problematic, color and padding need to be on the same element.
I thought about how to solve this, and I arrived at the conclusion, that, since this is the design system, there will be some common styling to the children, like background color of the box.

frontend/packages/design-system/src/components/Accordion/index.ts
internetti: yeah fair enough. not really useful here. could still change it

PR #385, opened 2022-03-07
Cor 492 read subrecord values backend

neb42:
* Extend `FieldValue.Value` to accept `[][]FieldValue`
* When reading records, also read sub records and populate the response

TODO:
* Tests
* Rename `StringOrArray`

Future work:
* Take a similar approach to writing records. I think we currently have to make a POST request for every subrecord we create rather than a single payload.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,14 +1,8 @@
 approvers:
   - ludydoo
   - internetti
-  - shinroo
-  - senyaoh
-  - nilueps
   - neb42
 reviewers:
   - ludydoo
   - internetti
-  - shinroo
-  - senyaoh
-  - nilueps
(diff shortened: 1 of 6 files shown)


internetti (commented):
definitely prefer this version!
but I just tried to create a record with multiple sub forms with multiple records to test this, and got an error from the backend, pretty cryptic, I think the relevant part is this `2022/03/07 16:09:12 runtime error: invalid memory address or nil pointer dereference`, seems to come from `readInSubFormField`, line 378
do you get this error too?

neb42:
@internetti I've added a short-term fix. I think the correct fix for this would be to create records and subrecords together

PR #386, opened 2022-03-09
COR-489: load recipient into profile page

internetti:
[COR-489]

requesting a recipient by the id in the url from the api and pass the data to the RecordView component


frontend/packages/core-api-client/src/client/Form.ts
```diff
@@ -32,4 +34,24 @@ export class FormClient {
...
+    const reference = formResponse.response.fields.find(
+      (field: FieldDefinition) => field.fieldType.reference,
+    );
+
+    if (reference && reference.fieldType.reference) {
+      const result = await this.getAncestors(
+        reference.fieldType.reference?.formId,
+      );
+      return [...result, formResponse.response];
+    }
```
neb42, line 54: I think the behaviour we want is slightly different. Rather than getting every reference field, we just want to get the reference fields that are keys.

frontend/packages/core-api-client/src/client/Recipient.ts
```diff
@@ -21,17 +31,66 @@ export class RecipientClient {
...
   };
 
-  get = async (args: RecordLookup): Promise<Recipient> => {
-    return Promise.resolve({
-      id: 'id',
-      formId: 'formId',
-      databaseId: 'databaseId',
-      ownerId: undefined,
-      values: [],
+  get = (args: RecordLookup): Promise<RecipientResponse> => {
```
neb42, line 38: Why have you changed the response type of this? I think we want to keep it as `Promise<Recipient>`

frontend/apps/core-app/src/screens/RecipientProfileScreen/RecipientProfileScreen.container.tsx
```diff
@@ -1,11 +1,80 @@
...
+              record: {
+                ...ancestors[1].record,
+                values: [
+                  ...ancestors[0].record.values,
+                  ...ancestors[1].record.values,
+                ],
+              },
+            },
+          ];
+          setData(mergedIndividual.concat(ancestors.slice(2)));
```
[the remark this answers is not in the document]
neb42: I suspect I may make some changes to this code in my ticket as I have to do similar logic, but I will also have to do the reverse when writing a form.

frontend/apps/core-app/src/screens/RecipientProfileScreen/__tests__/RecipientProfileScreen.component.test.tsx
```diff
@@ -1,10 +1,47 @@
+import { FormType } from 'core-api-client';
```
neb42, line 1: I'd probably mock out the `RecordView` component in this test

frontend/apps/core-app/src/screens/RecipientProfileScreen/RecipientProfileScreen.component.tsx
```diff
@@ -1,57 +1,45 @@
...
-      {form && record && <RecordView form={form} record={record} />}
-    </Styles.Container>
+    <ScrollView>
+      <Styles.Container>
+        {isLoading && <Skeleton h="20" p="4" />}
+        {error && (
+          <Text variant="heading" color="signalDanger">
+            {error}
+          </Text>
+        )}
```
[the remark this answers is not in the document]
internetti: visually, we'd probably use the notifications the designers are working on?

frontend/packages/core-api-client/src/client/Recipient.ts
internetti: moving the ancestor code into the get solved this though

internetti:
I'll add some tests

frontend/apps/core-app/src/screens/RecipientProfileScreen/RecipientProfileScreen.container.tsx
```diff
@@ -1,11 +1,93 @@
...
+      }
+      setIsLoading(false);
+    })();
+  }, [FORM_ID, DATABASE_ID, route.params.id]);
+
+  React.useEffect(() => {
+    if (data.length) {
+      setPrettifiedData(prettifyData(data));
+    }
+  }, [data]);
```
neb42, line 84: I don't think this should be state. If `prettifyData` is an expensive function this should be:
```
const prettifiedData = React.useMemo(() => prettifyData(data), [JSON.stringify(data)]);
```
But I would say it isn't expensive and this component does not rerender often so the overhead of `useMemo` is greater than just calling the function on every render.
```
const prettifiedData = prettifyData(data);
```

neb42, line 84: `[JSON.stringify(data)]` because the deps check here is shallow

frontend/packages/core-api-client/src/client/Form.ts
```diff
@@ -32,4 +34,22 @@ export class FormClient {
...
+
+    if (!formResponse.response) {
+      throw new Error(formResponse.error);
+    }
+
+    const key = formResponse.response.fields.find(
+      (field: FieldDefinition) => field.key,
+    );
+
+    if (key && key.fieldType.reference) {
```
neb42, line 49:
```
    const referenceKey = formResponse.response.fields.find(
      (field: FieldDefinition) => field.key && key.fieldType.reference,
    );

    if (referenceKey) {
```

internetti: kinda thought, since recipient are required to have exactly 1 key, which has to be a reference, this might be enough, but better save than sorry :heavy_check_mark:

PR #387, opened 2022-03-09
Add stubs for recipient registration child components

neb42:
Stubs for child components in the registration flow. Lots of files, but little content.


frontend/apps/core-app/src/screens/RecipientRegistrationScreen/RecipientRegistrationScreen.container.tsx
```diff
@@ -1,11 +1,42 @@
...
+  React.useEffect(() => {
+    (async () => {
+      const formsResponse = await getForms();
+      setForms(formsResponse);
+    })();
+  }, []);
+
+  React.useEffect(() => {
+    setRecords(forms.map(buildDefaultRecord));
+    setMode('register');
```
internetti, line 23: is this necessary when it's the default?

neb42: I'm not sure what you mean

internetti: mode is by default set to register, and I don't see any point where it could be set to review, so this didn't seem necessary. but the code setting mode to review will come later, and then this will make sense, so nevermind.

internetti:
snapshots need to be updated

PR #389, opened 2022-03-14
Expo fixes

neb42:
A few fixes that I found when trying to get android auth working


internetti (commented):
do I have to set something locally?
I cannot start the app, neither the web version, nor the android version, neither on my device, nor in the emulator.

running yarn start results in the error `Error loading DevTools - No scheme specified for development client`

neb42:
> do I have to set something locally?
I cannot start the app, neither the web version, nor the android version, neither on my device, nor in the emulator.
 running yarn start results in the error Error loading DevTools - No scheme specified for development client

Hmm web should be fine. There's a bit of setup for android that's needed, can run through that on slack if you want but android won't run much right now.

internetti:
still have not found a solution, but I have the same issues on main, so this PR does not seem to be the problem

/lgtm

PR #394, opened 2022-03-24
Cor 486 registration page save

neb42:
* Add client method to create a record and it's subrecords
* Add a client method to create all records for a recipient
* Hookup saving recipient
* Rearrange navigation so all of the recipient routes are in the same stack
* Fix a bug with ReactHookTransformer not handling subforms properly
* Fix a bug where the subform modal was not reset when closing

TODO
* Tests

### DIFF
--- a/frontend/apps/core-app/src/navigation/root.tsx
+++ b/frontend/apps/core-app/src/navigation/root.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import { createDrawerNavigator } from '@react-navigation/drawer';
 
-import { RecipientRegistrationScreen } from '../screens/RecipientRegistrationScreen';
 import { routes } from '../constants/routes';
 
 import { RecipientNavigator } from './recipients';
@@ -17,14 +16,6 @@ export const RootNavigator: React.FC = () => {
         options={{
           title: routes.recipientsRoot.pageTitle,
           headerTitle: routes.recipientsRoot.headerTitle,
-        }}
-      />
-      <Drawer.Screen
-        name={routes.recipientsRegistration.name}
-        component={RecipientRegistrationScreen}
-        options={{
-          title: routes.recipientsRegistration.pageTitle,
-          headerTitle: routes.recipientsRegistration.headerTitle,
         }}
       />
     </Drawer.Navigator>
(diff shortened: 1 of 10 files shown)


frontend/apps/core-app/src/navigation/root.tsx
```diff
@@ -1,7 +1,6 @@
 import * as React from 'react';
```
internetti, line 1: do we still need the drawer navigator if there's only one thing to click on?

neb42: I think we will need some sort of root navigator in the future so it's useful to keep here. Unsure wether it will be a drawer or not (on mobile it probably will be)

frontend/packages/core-api-client/src/client/Record.ts
```diff
@@ -11,16 +11,58 @@ import {
...
 import { BaseRESTClient } from './BaseRESTClient';
 import { FormClient } from './Form';
 
-export class RecordClient {
+export class RecordClient implements RecordClientDefinition {
   private restClient: BaseRESTClient;
 
   private formClient: FormClient;
 
+  static buildDefaultRecord = (form: FormDefinition): Record => ({
```
internetti, line 26: what's your reason for putting this here? to me, a client is purely for communication with the api.

frontend/packages/core-api-client/tsconfig.cjs.json
```diff
@@ -3,5 +3,10 @@
   "compilerOptions": {
     "module": "commonjs",
     "outDir": "lib/cjs",
-  }
+  },
+  "exclude": [
```
internetti, line 7: just for information, why did you move this to the 'child' tsconfigs? did something not work?

neb42: So I had an issue that vscode wasn't working well on the test files because they were excluded in the tsconfig. Removing the exlusion rule fixed the issue, but broke the build. So I moved it to the build specific configs.

frontend/packages/core-api-client/src/client/Record.ts
neb42: I think of the client more for interacting with the "model" rather than just the api. Things that are generic enough for the `Record` model I would add here. This seems like a useful method to have across applications.

For example, I wouldn't add the code to transform records to react-hook-forms as that is specific to the client implementation.

neb42:
@internetti Do you have any more changes for this pr or can it be merged?

internetti (approved):

PR #395, opened 2022-03-28
COR-487: search page

internetti:
[COR-487]

list all recipients in table on search page.

used the Recipient.get function in the Recipient.list function, to be able to show the names in the table. turned off the fetching of the subforms for that usecase.


frontend/apps/core-app/src/screens/RecipientListScreen/RecipientListScreen.container.tsx
```diff
@@ -23,18 +25,34 @@ export const RecipientListScreenContainer: React.FC = () => {
   const [tableInstance, setTableInstance] =
     React.useState<SortedFilteredTable<RecipientListTableEntry> | null>(null);
 
+  const [_, formState] = useAPICall(
+    formsClient.Form.getAncestors,
```
internetti, line 29: this should not be getAncestors, might use the leaf node approach for delivering the correct forms here

frontend/apps/core-app/src/components/RecipientListTable/RecipientListTableContainer.tsx
```diff
@@ -10,11 +10,13 @@ import { RecipientListTableEntry, SortedFilteredTable } from './types';
 
 type Props<T extends Record> = {
   data: FormWithRecord<T>[][];
+  form: FormDefinition;
```
neb42, line 13: I don't think this needs to be passed in. We can calculate the columns from `FormWithRecord`

neb42 (commented):
I think the data loading and table data is wrong here. You're currently getting the ancestors of the leaf node and showing a table for each (with the wrong fields for them). I think it'll be best to merge in the leaf node pr and then change it to do this:

1. Read leaf node forms from context
2. Select first leaf node from list (temporary until we have a UI to choose between leaf nodes)
3. List records for leaf node
4. Render a single table for leaf node

----------

We can also probably look at rendering less table columns. Maybe something like 5 columns, starting at the root of `FormWithRecord` and working down.

----------

I don't think the table is growing to the full width. If you use a form with less columns they wont grow to fit the space.

frontend/apps/core-app/src/components/RecipientListTable/RecipientListTableContainer.tsx
neb42: For that state I'd typically replace the table with some form of placeholder text (the subform table does something like this)

frontend/packages/core-api-client/src/client/Recipient.ts
```diff
@@ -64,8 +64,23 @@ export class RecipientClient implements RecipientClientDefinition {
     );
   };
 
-  public list = (): Promise<RecipientList> => {
-    return Promise.resolve({ items: [] });
+  public list = async ({
+    formId,
+    databaseId,
+  }: FormLookup): Promise<FormWithRecord<Recipient>[][]> => {
```
neb42, line 70: I think this change is worth doing outside of this pr as a refactor later.

I'm thinking `FormWithRecord<Recipient>[][]` isn't the best return type here as we're duplicating a lot of data as the forms are the same for each record. We could probably reduce the api calls made by doing a `list` for each form and then connecting the records together.

Something like:
```
{
  forms: FormDefinition[];
  records: Record<string, Record>[]; // Different Record types
}
```
or
```
{
  forms: FormDefinition[];
  records: Record[][];
}
```

or maybe `FormWithRecord<Recipient>[][]` is the best return type as it's used by the frontend widely, but how we build it could be changed.

internetti: yeah, the current structure is definitely not ideal, we should do this when we find time

neb42 (commented):
I haven't tested this yet as I'm having local tunnel issues

frontend/apps/core-app/src/components/RecipientListTable/createTableColumns.ts
```diff
@@ -1,22 +1,31 @@
-import { FormWithRecord, Record } from 'core-api-client';
+import { FormWithRecord } from 'core-api-client';
 import { Column } from 'react-table';
+import { Recipient } from 'core-api-client/src/types/client/Recipient';
 
 import { RecipientListTableEntry } from './types';
 
 export const createTableColumns = (
-  data: FormWithRecord<Record>[][],
+  recipientWithForm: FormWithRecord<Recipient>[] | null,
```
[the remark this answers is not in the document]
internetti: because I need all of the forms of one recipient, this was the easiest way to get them

PR #396, opened 2022-03-29
Cor 486 registration page validation

neb42:
Adds form validation and validates the recipient hierarchy when creating.

I want to look into using io-ts for validation (there's a plugin for rhf) but this will suffice for now.

### DIFF
--- a/frontend/packages/core-api-client/src/validation/Recipient.ts
+++ b/frontend/packages/core-api-client/src/validation/Recipient.ts
@@ -1,0 +1,30 @@
+import { FieldKind, FormWithRecord, Record } from '../types';
+import { getFieldKind } from '../utils';
+
+export const validateRecipientHierarch = (
+  formWithRecordList: FormWithRecord<Record>[],
+): void => {
+  formWithRecordList.forEach(({ form, record }, idx, arr) => {
+    if (form.id !== record.formId) throw new Error();
+
+    if (idx === 0) {
+      // assert no key field
+      form.fields.find((f) => f.key);
+    }
+
+    // assert reference key field
+    const keyField = form.fields.find(
+      (f) => f.key && getFieldKind(f.fieldType) === FieldKind.Reference,
+    );
+
+    if (!keyField) throw new Error();
+
+    const { form: prevForm } = arr[idx - 1];
+
+    if (
+      keyField.fieldType.reference?.formId !== prevForm.id ||
+      keyField.fieldType.reference?.databaseId !== prevForm.databaseId
+    )
+      throw new Error();
+  });
+};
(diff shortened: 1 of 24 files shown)


neb42:
There's also probably some more validation rules we can add, but this sets up a baseline

frontend/packages/core-api-client/src/validation/Recipient.ts
```diff
@@ -0,0 +1,30 @@
+import { FieldKind, FormWithRecord, Record } from '../types';
+import { getFieldKind } from '../utils';
+
+export const validateRecipientHierarch = (
```
internetti, line 4:
```suggestion
export const validateRecipientHierarchy = (
```
was this a typo?

frontend/packages/core-api-client/src/validation/Recipient.ts
```diff
@@ -0,0 +1,30 @@
...
+  formWithRecordList: FormWithRecord<Record>[],
+): void => {
+  formWithRecordList.forEach(({ form, record }, idx, arr) => {
+    if (form.id !== record.formId) throw new Error();
```
internetti, line 8: shouldn't we pass some kind of message with the errors?

frontend/packages/core-api-client/src/validation/Recipient.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    if (idx === 0) {
+      // assert no key field
+      if (form.fields.find((f) => f.key)) throw new Error();
```
internetti, line 12: this is about checking if this is the root individual right? why shouldn't the root have a key? first and last name together could be a key. shouldn't we rather check for reference here?

frontend/packages/core-api-client/src/validation/Recipient.ts
neb42: Yeah good catch. I'll do it in a commit rather than the suggestion as I'll need to change it's uses.

frontend/packages/core-api-client/src/validation/Recipient.ts
neb42: Yeah, I think I need to add something for displaying api errors too

frontend/packages/core-api-client/src/validation/Recipient.ts
internetti: typo is till there ;)

frontend/packages/core-api-client/src/validation/Recipient.ts
internetti: so what happened to passing error messages here?

PR #401, opened 2022-04-26
More auth fixes

neb42:
This fixes an issue where all tokens had expired and the UI wouldn't fallback to the login screen

### DIFF
--- a/frontend/apps/core-app/src/components/AuthWrapper/useTokenResponse.ts
+++ b/frontend/apps/core-app/src/components/AuthWrapper/useTokenResponse.ts
@@ -135,7 +135,11 @@ export const useTokenResponse = (): [
       if (refreshTokenInterval.current)
         clearInterval(refreshTokenInterval.current);
     };
-  }, [tokenResponse?.refreshToken, tokenResponse?.expiresIn]);
+  }, [
+    tokenResponse?.refreshToken,
+    tokenResponse?.expiresIn,
+    JSON.stringify(discovery),
+  ]);
 
   React.useEffect(() => {
     if (tokenResponse) storeTokenResponse(tokenResponse);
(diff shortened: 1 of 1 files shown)


frontend/apps/core-app/src/components/AuthWrapper/useTokenResponse.ts
```diff
@@ -27,7 +27,11 @@ const getStoredTokenResponse = (): TokenResponse | undefined => {
 const storeTokenResponse = (tokenResponse: TokenResponse) => {
   try {
     if (Platform.OS !== 'web') return;
-    sessionStorage.setItem('tokenResponse', JSON.stringify(tokenResponse));
+    if (tokenResponse) {
```
internetti, line 30: might be useful to write a test for this

frontend/apps/core-app/src/components/AuthWrapper/useTokenResponse.ts
```diff
@@ -131,7 +135,11 @@ export const useTokenResponse = (): [
       if (refreshTokenInterval.current)
         clearInterval(refreshTokenInterval.current);
     };
-  }, [tokenResponse?.refreshToken, tokenResponse?.expiresIn]);
+  }, [
+    tokenResponse?.refreshToken,
+    tokenResponse?.expiresIn,
+    JSON.stringify(discovery),
```
ludydoo, line 141: That might be expensive to stringify for change detection. Would this work with `discovery` alone ?

neb42: Ran a test for it and it's below a ms

frontend/apps/core-app/src/components/AuthWrapper/useTokenResponse.ts
internetti: @neb42 are you planning to write a test here, or do you consider this PR done?

neb42: Having trouble mocking session storage. I think the logic here is simple enough to leave it

PR #405, opened 2022-05-09
Core db stubs and create flow

neb42:
Opening this early so people can have a look. It probably shouldn't be merged until we're happy with the architecture.

* New layout for microservice code. Here the code related to core db is entirely isolated from the rest of the code base (save for some of the deployment commands/config). The only code I've added outside of this directory is for common use across services. Code is also split out by resource within the microservice for further isolation.
* I think the main improvement over our current flows is that the previous store/sqlmanager/sql layers have been moved to service/model/sqlbuilder layers. I feel each layer here has a clearer responsibility. Before the sqlmanager was too tied to the business logic and cut across resources.
* `validation.Validate` to make validation more consistent and simple
* Stubs for unimplemented flows. This is something we should do going forward. It will make parallelising on the backend much quicker if we can define the flows before hand and implement each stub one by one.
* Added migrations outside of gorm. This uses [migrate](https://github.com/golang-migrate/migrate) with migrations written in raw sql.

One thing I noticed is that creating a new microservice touches a lot of places. This should either be improved or documented.

I initially wanted to do just the stubs here, but I got carried away and started the create flow as an example. It is probably 70-80% finished.
TODO:
- Finish the flow (eg. creating the record table)
- Validation + JSON marsheling/unmarsheling for each struct that needs it
- Tests


pkg/api/types/pagination.go
```diff
@@ -0,0 +1,5 @@
+package types
+
+type PaginatedData[T any] struct {
+	Data []T
```
internetti, line 4: shouldn't this have info on start and end index of the contained data? or page number? some meta data?

neb42: Yeah it's not complete. I was just messing around with generics. I'd probably flesh this out when doing the list endpoint

pkg/server/core-db/migrations/1_entity.up.sql
```diff
@@ -0,0 +1,43 @@
+BEGIN;
```
internetti, line 1: if we change things later, we'd add another file for the updates in this folder, right?

pkg/server/core-db/models/entity/insertAttribute.go
```diff
@@ -0,0 +1,60 @@
...
+			"entity_id",
+			"constraint_required",
+			"constraint_unique",
+			"constraint_min",
+			"constraint_max",
+			"constraint_pattern",
+			"constraint_enum",
+			"constraint_custom",
+		},
+		[]any{
```
internetti, line 37: could this be typed more specifically?

pkg/server/core-db/types/Relationship.go
```diff
@@ -0,0 +1,19 @@
...
+	OneToMany  Cardinality = "OneToMany"
+	ManyToOne  Cardinality = "ManyToOne"
+	ManyToMany Cardinality = "ManyToMany"
+)
+
+type EntityRelationship struct {
+	ID                string      `json:"id"`
+	Cardinality       Cardinality `json:"cardinality"`
+	SourceEntityID    string      `json:"sourceEntityId"`
+	SourceAttributeID string      `json:"sourceAttributeId"`
```
internetti, line 16: didn't see this in the frontend, what's this needed for?

pkg/sqlmanager/sqlbuilder.go
```diff
@@ -0,0 +1,54 @@
...
+	return sqlBuilder{}
+}
+
+func (s sqlBuilder) InsertRow(
+	schemaName string,
+	tableName string,
+	columns []string,
+	values []interface{},
+) schema.DDL {
+	var placeholders = make([]string, len(columns))
```
internetti, line 33: "placeholders" as in "defaults"

pkg/server/core-db/types/Relationship.go
neb42: I didn't do any relationships in the frontend client as I was only creating a single entity. It's used for defining relationships between entities (see the confluence docs for more details).

pkg/sqlmanager/sqlbuilder.go
neb42: This is a copy+paste of another function that was unfortunately made private

pkg/server/core-db/types/Relationship.go
internetti: yes, the relationship I was aware of, but why do the relationships need attributes? what's that for? and the relationship type in the frontend doesn't have these properties

neb42: Yeah I want to think through relationships a bit more. I've added a line in the confluence doc that the relationship stuff may be wrong. Previously I was thinking of relationships as being defined by the primary keys, but I was thinking here that they could also be defined on an attribute level. I'm actually leaning back towards doing it based on ids, but I want to spend more time thinking through this again.

As I was typing this one other thought I had, if we were to have a "question bank" then if entities pulled attributes from the question bank we would actually have some implicit relationships. Maybe not useful functionally, but something interesting to see on the analytics side.

pkg/sqlmanager/sqlbuilder.go
internetti: so, are they meant to be defaults or not?

ludydoo:
Boy.. this is a massive PR

pkg/common/transaction.go
```diff
@@ -0,0 +1,38 @@
...
+	"gorm.io/gorm"
+)
+
+type TransactionManager interface {
+	Begin() *gorm.DB
+	Commit(tx *gorm.DB) *gorm.DB
+	Rollback(tx *gorm.DB) *gorm.DB
+}
+
+type transactionManager struct {
```
ludydoo, line 14: Not sure about the use of `TransactionManager` and `transactionManager`... We already have these methods on the `db` object (either Gorm or SQLx or sql.DB)

docs/backend_standards.md
```diff
@@ -0,0 +1,200 @@
...
+      migrations/
+        1_description.up.go
+        1_description.down.go
+        2_description.up.go
+        2_description.down.go
+      types/
+      server.go
+```
+
+### Controllers
```
ludydoo, line 42: Isn't that exactly what the current `handlers` do ?

pkg/api/types/pagination.go
ludydoo: I would rather add metadata + interface to our current `Lists` objects rather than implementing another model collection

frontend/apps/core-app/src/clients/coreDBClient.ts
```diff
@@ -0,0 +1,90 @@
+/*
+  USED FOR TESTING PURPOSES ONLY
+*/
+
+import { BaseRESTClient } from 'core-api-client';
```
ludydoo, line 5: Not sure where to put this comment, but just based on naming (`code-DB-client`), we should probably be dealing with `tables`, `rows`, `fields`, `constraints` rather than `entities`

ludydoo: I think the advantage of having a low-level api such as a `DB` api would be to allow clients to interact with lower level concepts that will become necessary for reconciliation/offline use, like row versioning

ludydoo: I think the concept of `Entity` is too high-level for this kind of API, and we already have a some kind of a similar concept encapsulated in the `forms` api.

pkg/api/types/pagination.go
ludydoo: And returning a `next` bookmark in the metadata perhaps

pkg/server/core-db/models/migrate.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	driver, err := postgres.WithInstance(sqlDB, &postgres.Config{})
+
+	if err != nil {
+		return err
+	}
+
+	m, err := migrate.NewWithDatabaseInstance(
+		"file:///Users/bmcalindin/workspace/core/pkg/server/core-db/migrations",
```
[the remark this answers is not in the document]
ludydoo: We can embed the migrations with the (really good) https://pkg.go.dev/embed package

pkg/server/core-db/types/Relationship.go
ludydoo: I think that's gonna bite us in the ass. I would rather add low-level `FOREIGN KEY` indexes in `table` objects rather than trying to find an abstraction for those relationships

frontend/apps/core-app/src/clients/coreDBClient.ts
neb42: > we should probably be dealing with tables, rows, fields, constraints rather than entities

Happy to talk about naming of resources. I'm not set on anything, but I thought an abstraction over the specific database technology used would be good.

> we already have a some kind of a similar concept encapsulated in the forms api

What exactly are you referring to here?

docs/backend_standards.md
neb42: handlers = controller, sorry maybe I should have used handlers but I was just using the language I'm used to using. Happy with either

pkg/validation/validate.go
```diff
@@ -0,0 +1,31 @@
+package validation
+
+import "reflect"
+
+func Validate(s any, isNew bool) ErrorList {
```
ludydoo, line 5: I don't really like that objects have a `Validate` method. In my opinion this violates the **S**OLID (Single-Responsibility) principle. Validation logic should be in its own encapsulated function/component/class. That will become increasingly evident once validation logic needs to make database calls, and then we need to inject a database handle into each Data Transfer Object (dto).

neb42:
> Boy.. this is a massive PR

Haha sorry, got carried away. Happy to talk anyone through it in more detail if useful.

frontend/apps/core-app/src/clients/coreDBClient.ts
neb42: Also I think these conversations are better for the confluence docs. This pr is more about the go api structure. I'm just using core-db as a reason to do it.

ludydoo: Yeah sure, I think its really useful to have a concrete example such as this PR to have better idea of what are the general directions & to iterate on !

pkg/api/types/pagination.go
ludydoo: Such as 
```
type ListMetadata map[string]string
type RecordList struct {
  Metadata ListMetadata
  Items: ?
}
/////
{
  "metadata":{"next":"9e44c0273dc3b838"}
  "items":[
    {...}
  ]
}
```

pkg/common/transaction.go
ludydoo: I see, I think its a bit early to introduce this since the `sql.DB` is exclusively called from whatever `store` we have. We kind of don't need yet the handlers to have access to this interface. I think it's likely that it will remain like that, basically that the handler will simply `Unmarshal` `Validate` `Call Store.Something`, where the `Store` will do pretty much everything related to DB calls and so on

neb42: I think we definitely do need it. When you get to the service layer you'll see how.

```
Controller
  Unmarshal
  Validate
  Call Service.Create
Service
  TransactionManager.Begin
  Call Store.InsertA
  Call Store.InsertB
  Iterate over x:
     Call Store.InsertC
  TransactionManager.Commit
Controller
  Validate
  Marshal
```

pkg/api/types/pagination.go
neb42: I think the generic is much cleaner

frontend/apps/core-app/src/clients/coreDBClient.ts
ludydoo: What I mean by that is concerning the conversation we had about exposing low level concepts around `tables`, `records`, etc, its upcoming usefulness in terms of implementing offline, and that the `forms` api is basically syntactic sugar that makes it easy to create those complicated, low-level objects. The forms-api is a higher level concept that depends on the db-api. Wheras the db-api is basically the lowest level thing in core, and the most important one probably.

What I foresee is that a subset of this `core-db-api` will basically become our reconciliation endpoint for offline use. 

There are 2 apis I see in this microservice
1. Management api to create tables, constraints, stuff like that
2. Data api, to push data to those tables

The write part of (1) - Management API, should be exclusively for online use
The read part of (1) - Management Api, closely resemble what we need in terms of forms discovery for offline
The (2) - Data Api, closely resemble what we need in terms of apis for data reconciliation for offline

And whatever we come up with this Data Api will probably be what we need to implement on the client side as well, when we implement cross-device reconciliation. So this Data Api should be written in a way that it is agnostic to the fact that it is a server or a client or whatever else. 

We don't want the mobile devices or website to have a different api whether its talking to a server or to another client in p2p mode. We probably want this api to be exactly the same. So imho the most sensible part of this whole PR is exactly in the design of this data api.

pkg/common/transaction.go
ludydoo: But I don't see why we need a service layer ? This microservice "business task" is to talk to the database. So there is no need to encapsulate the database layer in an `infrastructure layer` (if you think DDD design). The database concepts should be first-class citizen of this whole microservice. What exactly would be the role of the `service` layer in that case ? I think the `core-db-api` is a really fancy `store`

frontend/apps/core-app/src/clients/coreDBClient.ts
neb42: Yeah I think I pretty much agree with you. It's the interface of the `CoreDBClient` that should stay the same between offline and online. Seen as the client basically just maps to api endpoints then the api interface should also be fairly consistent.

pkg/api/types/pagination.go
ludydoo: Yeah but in this case, we can't really use generics for each entity because we deal with runtime types. 
We sure can define a generic `ObjectList` such as
```
type ObjectList[T:Object] struct{
  Metadata: ListMetadata
  Items: []T
} 
```
and reuse that across

pkg/common/transaction.go
neb42: I think regardless of what we call them there is a need for 3 layers (after the controller):

1. Layer A - Orchestrating multiple database calls (service)
2. Layer B - Executing a single database call (store)
3. Layer C - Building database statements (sql builder)

pkg/api/types/pagination.go
neb42: Isn't that just what I had? (with the metadata)

pkg/common/transaction.go
ludydoo: I disagree. In all layers there is the same concept which is calling the database. These seem to me that they are just 3 distinct functions in the same layer rather than their own package with their own concepts. 
I think we do need some tooling to craft sql queries. But basically the problem seems to be

1. convert whatever payload to a series of operations
2. do sheanigans for reconciliation
3. use a DAG to sort the operations by dependencies
4. start transaction
5. perform each operation
6. end transaction
7. report

I really think we should stick to the most simple approach, since whatever we implement in this layer we basically have to fully and accurately replicate on the client side in Typescript. This part of the code should be able to run in client or server mode. If we distribute this in a N-Layered approach, then we need the same N-Layered approach for the client side as well. I honestly think we don't need it. It's going to be dense for sure, but this will be the logic for our core database engine for offline use as well.

pkg/api/types/pagination.go
ludydoo: I'm just seeing a `[]Data` field

ludydoo: Oh, lol yeah I see. Yes

pkg/common/transaction.go
ludydoo: One option we could investigate would be a program that 
1. Accepts records as inputs
2. Outputs SQL statements, expects those statements to be executed

So basically
```
1. client send RECORDS to PROGRAM
2. PROGRAM outputs SQL
3. client executes SQL and sends results to PROGRAM
4. ... 
```

In this way, this program would simply accept inputs and outputs the necessary SQL statements that need to be executed to accurately reconcile the records.

```
# pseudocode

engine := newEngine("postgres")
engine.OnSQL(func (sqlStatement string) (interface{}, error){
  fmt.Println("Engine wants to execute statement:", sqlStatement)
  return db.Exec(sqlStatement)
}}
engine.PutRecords(records)
```

pkg/sqlmanager/sqlbuilder.go
neb42: I think what it does is make `$1, $2, $3, etc` which correspond to `Args:  values,`. So when the statement is executed `values[0]` is `$1`

neb42: So not defaults

pkg/common/transaction.go
ludydoo: Basically a program that can maintain our invariants by converting a set of RECORD INSTRUCTIONS to SQL

ludydoo: Our `handlers` would basically just be exposing this engine over HTTP

neb42: I get your point about the service essentially being the store, but I don't like this `engine` api.

> 1. convert whatever payload to a series of operations
> 2. do sheanigans for reconciliation
> 3. use a DAG to sort the operations by dependencies
To me this is the service layer. When creating a REST resource we are actually creating multiple database resources. By putting this in the store we have to replicate it for whatever db we use. This belongs in a different layer that interacts with the store through an abstract interface.

Something like this could take away the tx/db dependancy from the service layer.
```
func (d *entityService) Create(ctx context.Context, entity types.Entity) (*types.Entity, error) {
	storeInstance := store.NewInstance()

	storeInstance.AddStatements(store.InsertEntityDefinition())

	for i, attribute := range entity.Attributes {
		store.AddStatements(store.InsertAttribute())
	}

	for i, relationship := range entity.Relationships {
		store.AddStatements(store.InsertEntityRelationship())
	}

	store.ExecuteStatements()
}
``` 
`storeInstance` keeps a list of statements then `ExecuteStatements` runs them sequentially within a transaction.

We could move the `Service.Create` to `Store.Create` and make the other store methods private, but I think this is overloading the store.

I think having this layered approach is actually simpler. Each layer has a clearer responsibility and is easier to test. Layers or not, all of the logic stays the same and we'd have to replicate it anyway.

neb42: > whatever we implement in this layer we basically have to fully and accurately replicate on the client side in Typescript
Build the api in typescript so we can reuse it?

frontend/apps/core-app/src/clients/coreDBClient.ts
neb42: > tables, rows, fields, constraints rather than entities

When you talk about the naming here, do you mean we keep the same data structures but rename them or are you talking about changing the data structures?

pkg/common/transaction.go
ludydoo: I would say yes in the Java, .Net, Python, Ruby, dot net or other programming paradigms where the frameworks encourage you to use dependency injection containers like Spring, Windsor & al., automatic field injection and so on. But the advantage and strength of go is that it provides enough out of the standard library to create an application that is almost as fast as raw C++, supports extreme parallelization, etc. without needing any third party library. It is MUCH faster than NodeJs, Python, and others. It is also much easier to reason about because there is no framework-specific, automagical, convention-over-configuration stuff. It's all there in the code. There's also no annotation-based logic, such as NestJS @Controller and others.

Another advantage is portability, since you can compile a go program into a portable executable that runs on all platforms, x86, windows, ios, raspberry pies, wasm, etc. You can end up with a container that contains a single file, the executable itself, which is a huge win for security, performance, and eventually cost. The recurring problem with NodeJS, python & al, is that you end up with thousands of dependencies, direct or indirect. And it makes it extremely difficult to bundle, and very long to build. By removing these dependencies, you also remove the need to keep up with vulns that are spread over thousands of libraries, and have to manage a very limited subset instead. NodeJS's founder even publicly said that he would rather use Go (or deno) for distributed apis rather than Node. 

Unless there is a library that absolutely requires it, I don't see any reason not to use go.

And in the microservice architecture we are leaning towards, the service is the program. I think there would need to be a very specific usecase if we were to use services within microservices. For example, in a monolith app, we would probably have some kind of a `CoreDbService` or something like that. But with microservices, we already made that cut upstream, and divided code so that `CoreDbService` is its own microservice. So it seems to me that introducing a N-Layered architecture within a microservice architecture is overkill.

What we are building here is not a service per-se. We're more like coming up with a protocol, and programming a node to be part of a network of peer-to-peer clients, so that the data is available at all times. I think that's the main shift in how we think about this project vs other regular web application projects. This is not a web application. Given a node in the network, it would see the address of this server like the address of any other probably

```
# peers.ts
{"peers": [10.22.104.203, 11.103.34.21, ...]}
```

Thie `core-db` resembles more a storage node than a web application. A node that only provides the Durability guarantee of a distributed, p2p storage solution. Which in a way, would perhaps make more sense if it was a Pull-based system rather than a Push-based system. That would probably make it much easier for the reconciliation logic, as each actor in the network would be responsible to keep its data up to date, rather than expecting other actors to push data to it. If each actor exposes a read api, it is sufficient for each actor to keep things up to date within themselves. In a way, there is no need for a public write api. Each actor could poll other actors in the network in a pull-based fashion, rather than having to deal with potential concurrent pushes to itself.

We could use something like the `gossip` protocol to distribute the latest timestamp of the latest change etc. between all actors in the network. Those could choose to pull the data when they see fit. If there is a discrepancy between the lastest local pulled state vs the gossiped latest state, then the actor knows it should pull stuff.

frontend/apps/core-app/src/clients/coreDBClient.ts
ludydoo: I think it makes sense to have a DB service, but the entities within this service should be form the DB domain, such as `tables` `fields` `constraints` etc. rather than `entities`

pkg/common/transaction.go
ludydoo: Advantage of gossip is that it will support fragmented networks, such as when internet goes down, and is extremely fast

frontend/apps/core-app/src/clients/coreDBClient.ts
neb42: Do you mean that for the api interface rather than `GET /entity`, you would have:
```
GET /table
GET /field
GET /constraint
```
or would the `table` entity still encompass the `fields` and `constraints`?

pkg/common/transaction.go
neb42: I'm not really advocating for typescript as I agree with most of your points about go. The main benefit it would have is development speed and code sharing with the frontend. I think it's not the best choice, but probably the more pragmatic choice given a team of our size. Again, not suggesting we switch.

I don't agree with your point about not using dependency injection in go. Sure it may not explicitly encourage it, but it's still a good pattern to use. For example when we switch from writing to standard sql tables to writing to an event stream, we can easily swap out a single module.

I don't think a microservice architecture is a reason not to use a layered architecture within the service. If a service requires it then we should at it. Otherwise it can easily become very bloated and difficult to work with.

On the service layer, I think there are other use cases for it. In the new forms api it will need to communicate with the core-db api. I suspect this will warrant a service layer, although I haven't thought through this in detail yet.

On the distributed stuff. Yeah that all sounds good (without thinking too much on it), but I think we need to be pragmatic and consider this as a web service for now. If we start trying to design this as a p2p system without actually making a p2p system we'll end up down a rabbit hole and nothing will get done.

neb42: I've pushed up an example to `models/foo` that removes the service layer. I don't think it's right as the logic in `Create` is generic enough to be removed from the postgres implementation. But I thought it's worth having an example there to discuss.

caparisi (approved):

PR #406, opened 2022-05-13
replication poc

ludydoo:


frontend/apps/db-explorer/src/client.ts
```diff
@@ -0,0 +1,166 @@
...
+}
+
+export type GetRecordsRequest = {
+  table: string;
+  recordIds?: string[];
+  revisions?: boolean;
+}
+
+export type PutRecordRequest = {
+  isReplication?: boolean;
```
internetti, line 42: I'm a bit confused about the terms revision/replication? do you use them interchangeably or is there a difference in meaning?

internetti: also, what problem was solved by adding this? why do we need to differentiate between original and copy? is this just a shortcut to not have to check for revisions every time?

ludydoo: It's mostly due to (when we enable offline), that multiple people might edit the same records. When that happens, it's unclear whose version of the record is "the good one". To "solve" this, we store all the edits as different "revisions". Then, the engine chooses one of the revisions as being the "good one", but we can still present to the user that there was a conflict, and the user could resolve the conflict by choosing the "right one" manually. 

It's also needed for replication/synchronization between different devices. Basically, the devices would pull the revisions of the records that they don't know about. And since the process of "choosing the good one" is deterministic, all users will always elect the same revision all the time.

ludydoo: In this case `isReplication` is exclusively used by the replicator. The engine assumes that the revision being `put` is known by another `engine`.

pkg/server/data/api/revision.go
```diff
@@ -0,0 +1,109 @@
...
+	// The length of the hash is 32 (md5)
+	// The length of the separator is 1
+	// The length of the number is at least 1
+	// so the minimum length is 34
+	if len(str) < 34 {
+		return rev, NewError(ErrCodeInvalidRevision, "Invalid revision length")
+	}
+	hash := str[len(str)-32:]
+	// the hash can only contain hex characters
+	for _, c := range hash {
```
internetti, line 80: couldn't we do this with a regex?
specifially the hex check, but also the length checks?

ludydoo: We could for sure, but this is probably way more performant than compiling + executing a regex! That will add up if there is a lot of traffic. Also I tried to make this package depend on as little libraries as possible.

pkg/server/data/client/client_http.go
```diff
@@ -0,0 +1,114 @@
...
+
+type HTTPClient interface {
+	CreateTable(ctx context.Context, request api.Table) (api.Table, error)
+	GetRecord(ctx context.Context, request api.GetRecordRequest) (api.Record, error)
+	GetRecords(ctx context.Context, request api.GetRecordsRequest) (api.RecordList, error)
+	GetTables(ctx context.Context, request api.GetTablesRequest) (api.GetTablesResponse, error)
+	GetTable(ctx context.Context, request api.GetTableRequest) (api.Table, error)
+	GetChanges(ctx context.Context, request api.GetChangesRequest) (api.Changes, error)
+	PutRecord(ctx context.Context, request api.PutRecordRequest) (api.Record, error)
+}
```
internetti, line 32: it seems like there's no method to create a new record? am I missing something, or is there a trick?

ludydoo: `PutRecord` ;)

internetti: ah, the id in the request irritated me

pkg/server/data/engine/pkg.go
```diff
@@ -0,0 +1,47 @@
...
+// The engine is designed to be used in standalone mode or as a part of a
+// peer-to-peer network.
+//
+// # Introduction
+// The engine is designed to store and retrieve records in a table.
+// A table is a collection of records.
+// A record is a collection of fields.
+// A record has a unique id.
+// A field is a named value.
+// A value is a string, a number, or a boolean.
```
internetti, line 16: no date/time fields?

ludydoo: Not yet! This was just for POC...

pkg/server/data/engine/pkg.go
```diff
@@ -0,0 +1,47 @@
...
+//
+// When multiple versions of a record are in a conflict, the engine will
+// choose the version with the highest revision number. This is the version
+// of the record that will be reflected in the <table> table.
+//
+// The revision of a record is a string in the format <num>-<hash>.
+// The <num> is an incrementing number.
+// The <hash> is a hash of the record.
+//
+// Two records might have the same number but different hashes.
```
internetti, line 40: if the hash is of the original record, how can two revisions of the same record have a different hash?

internetti: or, if the hash is of the new version of the record, I don't understand yet how the revisions are linked to their original record

ludydoo: let's say you and I checkout Record1 with hash `1-aaa`. We go offline. We both update the record. 

I created a new revision with hash `2-nnn` and you `2-mmm`. Both records started at `1-aaa`, and they are both at the "2nd" revision, but they have different hashes because their content is different.

ludydoo: If we would for some reason update the same record `1-aaa` with the same contents on both of our sides, then we would compute the same hash, so that would be the same revision. (both you and i would get like `2-ggg` and `2-ggg`)

ludydoo: like
```
# we checkout this record
{"id":"some_id", "revision":"1-aaa", "attributes: {"foo":"bar"}}
# we go offline

# you update your record to have {"foo":"baz"}, so your store contains
{"id":"some_id", "revision":"1-aaa", "attributes: {"foo":"bar"}}
{"id":"some_id", "revision":"2-bbb", "attributes: {"foo":"baz"}}

# i update my record to have {"foo":"snip"} so my store contains
{"id":"some_id", "revision":"1-aaa", "attributes: {"foo":"bar"}}
{"id":"some_id", "revision":"2-ccc", "attributes: {"foo":"snip"}}

# we go online
# we reconcile with the server, so the server contains
{"id":"some_id", "revision":"1-aaa", "attributes: {"foo":"bar"}}
{"id":"some_id", "revision":"2-bbb", "attributes: {"foo":"baz"}}
{"id":"some_id", "revision":"2-ccc", "attributes: {"foo":"snip"}}

```

internetti: thanks for the example. the concept with the alphabetical order was quite clear, I just didn't realize that the revision is a property on the record. I thought the 2-mmm was supposed to somehow lead to the record itself, so the hash changing confused me. but if the revision is a property on the record, the hash is not relevant for the link.

neb42:
My understanding of the roadmap is that we had decided to do a centralised/online only version of core. The majority of this POC seems to be focussed on the concepts of revisions, p2p, and reconciliation.

Our goal at the moment is to design our data model and based on this POC I'm not sure how to progress with that goal.

I guess I'm not too sure of your aim of this pr. Are you suggesting we should be building something along these lines for the architecture refactor or is this more to demonstrate your ideas for work in the further future?

I do appreciate the point of having a single entry point of an `engine` that can be put into any use case, but I think that is doesn't depend on the p2p/reconciliation. I also think the interface of the engine is a bit overloaded and could be rethought.

I'd also like to revisit the pros/cons of using a revision based approach vs a crdt system.

neb42:
Something like this would separate the concerns of the engine a bit better

```
package engine

import (
	"context"

	"github.com/nrc-no/core/pkg/server/core-db/types"
)

type TableReader interface {
	Get(ctx context.Context, tableID string) (*types.Table, error)
	List(ctx context.Context) ([]types.Table, error)
}

type TableWriter interface {
	Upsert(ctx context.Context, table types.Table) (*types.Table, error)
	Delete(ctx context.Context, tableID string) error
}

type RecordReader interface {
	Get(ctx context.Context, tableId string, recordID string) (*types.Record, error)
	List(ctx context.Context) ([]types.Record, error)
}

type RecordWriter interface {
	Upsert(ctx context.Context, record types.Record) (*types.Record, error)
	Delete(ctx context.Context, tableID string, recordID string) error
}

type ReaderEngine struct {
	Table  TableReader
	Record RecordReader
}

type WriterEngine struct {
	Table  TableWriter
	Record RecordWriter
}

type Engine struct {
	Reader ReaderEngine
	Writer WriterEngine
}

func main() {
	tableReader = NewPostgresTableReader()
	tableWriter = NewPostgresTableWriter()
	recordReader = NewPostgresRecordReader()
	recordWriter = NewPostgresRecordWriter()

	engine := Engine{
		Reader: ReaderEngine{
			Table:  tableReader,
			Record: recordReader,
		},
		Writer: WriterEngine{
			Table:  tableWriter,
			Record: recordWriter,
		},
	}

	engine.Reader.Table.Create(...)
}
```

neb42:
I was thinking about this a bit more and have realised that you're showing how we could do this without actually having a data model stored in the db.

The issues that come to mind with this approach:

* Strong lock-in with sql. Even when looking at sql flavours there could be a feature disparity (sqlite doesn't have everything postgres has)
* Difficult to query table metadata. We would need to parse a `DESCRIBE` query or use the metadata stored in sql and parse it into our data structures.
* How would we do access control here? We could use schemas in postgres and different database files in sqlite, but this limits us to a single project level of access.
* We lose custom metadata about tables, like who created it.

Seen as we already need to define a data model for the api, then storing this isn't that much extra and gives us some benefits.

PR #407, opened 2022-05-18
Cor 484 scopes n claims

internetti:


pkg/api/types/identity_provider.go
```diff
@@ -18,9 +25,71 @@ type IdentityProvider struct {
 	// TODO: add unique constraint for email domains
 	// TODO: add support for multiple email domains for a single IdentityProvider
 	EmailDomain string `json:"emailDomain"`
+
+	Scopes string `json:"scopes"`
+
+	Claim Claim `json:"claim"`
```
ludydoo, line 31: Rename to `ClaimMappings` ?

pkg/api/types/identity_provider.go
```diff
@@ -18,9 +25,71 @@ type IdentityProvider struct {
...
+}
+
+type Claim struct {
```
ludydoo, line 34: Also perhaps rename to `ClaimMappings`

pkg/api/types/identity_provider.go
```diff
@@ -18,9 +25,71 @@ type IdentityProvider struct {
...
+
+func (c Claim) Value() (driver.Value, error) {
+	j, err := json.Marshal(c)
+	return j, err
+}
+
+func (c *Claim) Scan(src interface{}) error {
+	source, ok := src.([]byte)
+	if !ok {
+		return errors.New("type assertion .([]byte) failed")
```
ludydoo, line 52: I would add a more helpful error message such as 
```
return fmt.Errorf("failed to scan scan claims: value of type %T could not be converted to []byte", src) 
```
that would also be helpful for the type assertions below

ludydoo:
I think that we do not need the `Scan` or `Value` methods in the `api` package, since those methods are only called by the database driver to convert to and from a value that comes from and to the database!

internetti:
the `Value` method seems to be needed there, if it's missing, the update fails with `cannot convert {6 map[a:b]} to Bytea`

internetti:
> I think that we do not need the Scan or Value methods in the api package, since those methods are only called by the database driver to convert to and from a value that comes from and to the database!

the Value method seems to be needed there, if it's missing, the update fails with cannot convert {6 map[a:b]} to Bytea

ludydoo:
I think that's because we don't convert from the `api` types to the `store` types... Because only 1 of these need to be converted to/from the database. I can check with you tomorrow perhaps ?

